PR #26, opened 2025-02-04
feat: link end user to accounting point and update authz (FLEX-448)

enzo-crance-statnett:
The goal of the ticket is to link the accounting point to the end user, so that we can give authorisations to the EU on resources related to the accounting point, and notify them on processes involving these resources.

IIUC, in our system, an end user is always a party, because they assume a party of type `end_user` to interact with the FIS. This allows us to make two simplifications :

1. Instead of storing a business ID for the end user, we can _uniquely_ identify them with a **party ID**. This removes a level of indirection and we can make sure they are an end user by checking the party type in the foreign key check.

2. Knowing whether the end user is a person or organisation **does not seem to be required** in this table for now. If it happens to be in the future, when we refine the information we store, this is likely to end up in the `party` resource later on, not in the accounting point, so the information will always remain reachable from the party ID.

If we receive external data to update the accounting point table linking accounting point IDs to end user business IDs, we will just have to close the gap between end user business ID and party ID before actually updating the table.


enzo-crance-statnett:
### History

The simple rule was that history is visible when the main resource is visible. But if the visibility of the main resource is based on a time-dependent field, when the resource is not visible anymore, maybe we want to still grant access to history on the period when the time-dependent field used to grant access to the main resource.

#### Example for CU

Before
> We can see CU history when we can see the CU.

After
> We can see CU history on periods where we could see the CU.

krilor:
> ### History
> 
> The simple rule was that history is visible when the main resource is visible. But if the visibility of the main resource is based on a time-dependent field, when the resource is not visible anymore, maybe we want to still grant access to history on the period when the time-dependent field used to grant access to the main resource.
> #### Example for CU
> 
> Before
> 
> > We can see CU history when we can see the CU.
> 
> After
> 
> > We can see CU history on periods where we could see the CU.

Yes. I think that makes sense. These time-based authorizations are a bit wonky to work with, but that is the general policy we are after.

enzo-crance-statnett:
> [!NOTE]
> CU-SO001 and CU-SP001 rules are time-dependent too, so we may need to change the history rules (CU-SO003 and CU-SP004 which behave as former CU-COM001) if we think SO and SP should see the history of CUs they have been linked to in the past.
>
> => When it is not needed to see the past, we can just simplify and hide it when the user can no longer see the main resource.

enzo-crance-statnett:
Thoughts about CU-SP.

Even though this resource is time-dependent, here we do not care about _valid time_. It is just regular 
data like the other fields here, because it is not involved in the current checks.

We only care about _record time_. The end user can see what _happens_ on the CU-SP resource at the periods when they owned the AP.

enzo-crance-statnett:
Second round.

CU is a resource with _no_ valid time, so we should simply see the versions that successively become _active while we own the AP_.

=> **see history of CU if record time overlaps with AP valid time**

CU-SP is a resource with valid time, so changes can concern the past or the future. In particular, a user can :
- at any time, correct former contracts on the CU from when we did not own the AP ;
- at any time, plan future contracts on the CU for a period where we will not own the AP anymore ;
- in the past, plan the contracts for us on the period we are expected to own the AP ;
- in the future, correct contracts on the period we owned the AP.

A simple approximation for now is that we should see the two last categories, because they set up or correct stuff on the period where we are concerned, but not the two first ones, because they handle resources on the AP on periods where we do not own it.

=> **see history of CU-SP if valid time overlaps with AP valid time**

If a correction is needed, it will be made in a later story.

krilor:
> Justification for the implementation of notification receiver queries :
> 
> When a CU is created, we need to notify the end user associated to the AP (say `A`). But if the event is processed later and the end user is not the same as at the event time (say `B`), we notify `B` because with the additional knowledge we have at the time of _processing_ the event, `A` is no longer the end user for this CU at the time of the event. (this explains why we don't use history tables)
> 
> Note
> 
> Writing this made me wonder why the time of the event has any business value there. Maybe we should just check who is the end user _when the event is processed_ and notify this one. The one at the time of the event may just be outdated too ?
> 
> Regarding CU-SP contracts, we are pretty sure the time of the event does not mean anything because what matters is the _valid time_. Indeed, what matters is to notify the end user that owns the AP at the period concerned by the contract, so by using its _valid time_. To know who owns the AP, we take the best knowledge we have available : the latest one. Hence no history table either here.

I agree "who is the end user when the event is processed and notify this one" makes sense to me. 👍

krilor:
> When adding notifications, it was preferable to make queries go through `sqlc`, so we needed to add a view for AP end user in the `api` schema. But as we want to keep it internal (_i.e._, not exposed to the API), it had to stay out of the OpenAPI specification which we use for generating the schema for `sqlc`. Hence the local additional version of `schema.sql`.

Not answering this questions but I wonder if it is possible to just use the actual schemas/views rather than generating the fake ones? Basically point sqlc at the db/api directory? This would still require a custom schema.sql in this case, but would remove the need to do the "fake stuff".

On that note and ref https://github.com/elhub/flex-information-system/blob/main/docs-dev/modular-monolith.md . I think we are missing an "internal" api schema. Ref https://github.com/PostgREST/postgrest/discussions/2999#discussioncomment-7592206

> api: This actually consists of three schemas **api** (internal), **extra** (for db-extra-search-path) and **exposed** (db-schemas).

_If_ it is possible to point sqlc directly at our schema definitions then having this split would allow us to do what we want. We would just create a view in the "internal" api schema. We generally also want to do all operations towards the database via views/functions.

backend/event/models.sql
```diff
@@ -44,20 +53,39 @@ WHERE cusph.controllable_unit_id = @resource_id
...
-SELECT connecting_system_operator_id
--- not using _history because CUO ID is stable and actually from accounting_point
-FROM controllable_unit cu
-WHERE cu.id = (
-    SELECT controllable_unit_id
-    FROM controllable_unit_service_provider
-    WHERE id = @resource_id
-);
+SELECT apeu.end_user_id
+FROM accounting_point_end_user AS apeu
```
krilor, line 75: This is a bit nitpicky, but this join would benefit from being turned upside down. Basically start with cusp, and join until you find the end user. This is the "logical" join order in my HEAD.
This also allows us to go it all in one query.

```
SELECT unnest(
    array_remove(array[cusp.service_provider_id, cu.connecting_system_operator_id,apeu.end_user_id], NULL)
)::bigint
FROM controllable_unit_service_provider cusp
INNER JOIN controllable_unit cu ON cu.id = cusp.controllable_unit_id
INNER JOIN accounting_point AS ap ON cu.accounting_point_id = ap.business_id
LEFT JOIN accounting_point_end_user AS apeu ON ap.id = apeu.accounting_point_id
AND apeu.valid_time_range @> @recorded_at::timestamptz
AND apeu.valid_time_range && tstzrange(cusp.valid_from, cusp.valid_to, '[)');
WHERE cusp.id = @resource_id
```

^^ I havent tested this at all, just copied bits and pieces from the PR to show what I mean. Mistakes are expected.

db/api/accounting_point_end_user.sql
```diff
@@ -0,0 +1,18 @@
+-- Manually managed file
```
krilor, line 1: We should ideally not have this view in the API schema. All things in the API schema will be exposed by PostgREST.
It is "handled" by the limited GRANTS, I guess, but we really need to get a better handle on this. Ref the other comment about internal and exernal api schemas.

db/test_data/test_data.sql
```diff
@@ -146,34 +146,49 @@ $$ LANGUAGE plpgsql SECURITY DEFINER VOLATILE;
...
+    ) RETURNING id INTO ap_id;
+
+    INSERT INTO flex.accounting_point_end_user (
+      accounting_point_id,
+      end_user_id,
+      valid_time_range
+    ) VALUES (
+      ap_id,
+      eu_id,
+      tstzrange(current_timestamp, null, '[)')
```
krilor, line 188: We should probably do something about the valid_time_range here. First off it should probably be "midnight aligned" (norwegian time).

For now, just set a fixed timestamp? Later on we should probably make a proper valid time timeline with two different end users.

krilor (commented):
We are on the right path here 👍

krilor (approved):

PR #31, opened 2025-02-10
feat(ui): generate field tooltips from resource API docs (FLEX-456)

enzo-crance-statnett:
This PR adds tooltips to the `Show` pages for every field documented in the resource API specification.

Screenshot 2025-02-10 161953


frontend/src/auth/FieldStack.tsx
enzo-crance-statnett, line 1: Adding this to `FieldStack` allowed editing only one place and touching all the `Show` pages for free, which is a good first solution. It is not ideal though because the code ends up in the `auth` subfolder whereas this is not auth-related.

krilor (approved):
This looks super nice. Exactly the kind of thing  I was hoping we could do for this story 👍

PR #32, opened 2025-02-11
feat(ui): add tooltips to input components (FLEX-456)

enzo-crance-statnett:
Kind of 80/20 solution. Not perfect but very simple.

Screenshot 2025-02-11 100648

Other options explored (but more complex and did not manage on the first try) :
- show the tooltip inside the input field, as a Material UI `InputAdornment` ;
- show the tooltip before or after the input field, on the same _horizontal_ line.

> [!NOTE]
> Side feature of this PR. Dividers between fields on show pages makes the UX of tooltips even better IMHO :
>
> Screenshot 2025-02-11 100922


krilor (approved):
Is it possible to force the tooltip to be on the same line as the label (e.g. by wrapping it in something)? The input pages has way too much whitespace already.

PR #33, opened 2025-02-11
feat(db): implement AP(EU) RLS policies with a security definer view (FLEX-445)

enzo-crance-statnett:
This PR reimplements the RLS using `accounting_point` and `accounting_point_end_user` with an internal security definer view. First experimentation before further refactoring in our database architecture.

> [!NOTE]
> The view is in a distinct `authz` folder with already some logic in the scripts to be able to declare several of them later on. The intent is to isolate such _internal views_ from both _internal tables_ in the `flex` schema and _exposed views_ in the `api` schema.
>
> However, as the added internal view is used to define RLS policies, it still technically belongs to the `flex` schema. We can decide to change that later, but currently the architecture creates a mutual dependency between the `flex` schema and the (potential, future) `authz` schema.

### DIFF
--- a/db/authz/controllable_unit_accounting_point_end_user.sql
+++ b/db/authz/controllable_unit_accounting_point_end_user.sql
@@ -1,0 +1,19 @@
+CREATE OR REPLACE VIEW controllable_unit_accounting_point_end_user AS (
+    SELECT
+        cu.id AS controllable_unit_id,
+        ap.id AS accounting_point_id,
+        ap.system_operator_id AS connecting_system_operator_id,
+        apeu.end_user_id,
+        apeu.valid_time_range AS end_user_valid_time_range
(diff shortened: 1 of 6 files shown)


db/authz/controllable_unit_accounting_point_end_user.sql
```diff
@@ -0,0 +1,19 @@
+CREATE OR REPLACE VIEW controllable_unit_accounting_point_end_user AS (
```
krilor, line 1: I dont think we have use for the accounting point bits of this view. What is the rationale of exposing it? A view can be thought of as "interfaces" and keeping them small is IMHO a good idea if we can.

krilor (approved):
Consider removing the sp bits of the view. Else i think this is a great start and i like the idea of separating out authz for clarity.

PR #35, opened 2025-02-12
feat(db): relation between CU and ES + BRP via AP

enzo-crance-statnett:
Same model as the `CU <-> EU` relation we already have.

No notifications for these users required yet in the processes, so nothing implemented here for now.


krilor (approved):
I like it.

One thing to note is that the relation for BRP is actually not directly with the AP, like ES/EU. BRP is related to the ES via the price area/market balancing area (MBA).

We might have a benefit of storing the relation AP<->BRP directly (its nice for events and authorization), but we must know that the mapping probably must be done by whatever syncs from Elhub (AP-ES)/eSett (ES-BRP).

PR #36, opened 2025-02-12
feat(backend): move URL rewrites from nginx to the backend (FLEX-454)

enzo-crance-statnett:
Attempt at moving URL rewrites for single-ID and history pages from nginx to the backend.

### DIFF
--- a/backend/data/data.go
+++ b/backend/data/data.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"regexp"
 	"strings"
 
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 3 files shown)


backend/data/data.go
```diff
@@ -39,9 +40,52 @@ func NewAPI(
...
+func (brw bodyResponseWriter) Write(b []byte) (int, error) {
+	return brw.body.Write(b) //nolint:wrapcheck
+}
+
 // PostgRESTHandler forwards the request to the PostgREST API.
-func (data *API) PostgRESTHandler(ctx *gin.Context) {
+func (data *API) PostgRESTHandler(ctx *gin.Context) { //nolint:funlen
+	// regexes for calls targeting a single ID and history pages, which are not
+	// valid formats in PostgREST
+	regexID := regexp.MustCompile("^/([a-z_]+)/([0-9]+)$")
```
krilor, line 61: While in nginx it makes sense to have two patterns for this, does it really make sense when we are writing "proper" code? The `/history` bit could be optional.

That would avoid the branching/testing multiple regexes, and the only conditional would be to check the match group for history?

ALSO: Note that the `/history` is only one of the "subresources" we said we would support (we just haven't updated the nginx config to do it it seams. Ref our docs at https://github.com/elhub/flex-information-system/blob/main/docs/api-design.md#hierarchies-via-redirects . This means that the last capture groups should probably be something similar to the first one (but optional)

backend/data/data.go
```diff
@@ -50,23 +94,49 @@ func (data *API) PostgRESTHandler(ctx *gin.Context) {
 		req.URL.Scheme = data.postgRESTURL.Scheme
 		req.URL.Host = data.postgRESTURL.Host
 		req.URL.Path = url
+		req.URL.RawQuery = query.Encode()
+	}
+
+	/*
+		Single-ID calls are supposed to return only one record, but the rewrites
```
krilor, line 101: That is what the Accept header is all about. Ref https://github.com/elhub/flex-information-system/blob/main/local/nginx/templates/flex.conf#L126 and https://github.com/elhub/flex-information-system/blob/main/local/nginx/templates/flex.conf#L98-L100

krilor (changes requested):

krilor (approved):

PR #41, opened 2025-02-13
feat(backend): merge nginx rules for auth and data APIs (FLEX-454)

enzo-crance-statnett:


local/nginx/templates/flex.conf
```diff
@@ -75,30 +75,26 @@ server {
...
-        set $prefer $http_prefer;
+    # Proper API endpoints
 
-        proxy_set_header Prefer $prefer;
-        proxy_set_header Accept $accept;
+    location ~ ^/(api|auth)/v0/ {
 
-        proxy_hide_header Content-Type;
-        add_header Content-Type 'application/json' always;
+        proxy_set_header Prefer $http_prefer;
```
krilor, line 96: These should probably not be here. Looks like remnants from the POST stuff recently removed.

local/nginx/templates/flex.conf
```diff
@@ -109,7 +105,7 @@ server {
         proxy_hide_header X-Powered-By;
         proxy_hide_header Server;
 
-        proxy_pass ${NGINX_UPSTREAM_BACKEND}/api/v0/;
+        proxy_pass ${NGINX_UPSTREAM_BACKEND};
 
         # We want response that matches the error_message schema in the OpenAPI spec
         proxy_intercept_errors on;
```
krilor, line 111: Error messe interception should be done in the backend now.

krilor (commented):
There are a few things more that can be deleted. I'll push an update to the PR.

PR #43, opened 2025-02-13
chore: make project compatible with macOS

enzo-crance-statnett:
In the context of open sourcing, I thought it would make sense to make our project work on macOS, given that a substantial share of developers use this OS, and both other major OSes (Windows and Linux) are already kind of supported with WSL.

This PR does not touch dev documentation yet, but brings a few benign changes to make things work on macOS.


docker-compose.yml
```diff
@@ -46,7 +46,7 @@ services:
     expose:
       - "7000"
     ports:
-      - "7000:7000"
+      - "7001:7000"
```
enzo-crance-statnett, line 49: [macOS uses port 7000](https://github.com/cookiecutter/cookiecutter-django/issues/3499) for something that users likely do not want to turn off.

justfile
```diff
@@ -426,7 +426,7 @@ permissions-to-md:
     for resource in $(find docs/resources/ -type f -not -name "index.md" -exec basename {} \; | cut -d. -f1); do
         echo ".. ${resource}"
 
-        sed -i '/^For party type abbreviations/,$d' docs/resources/${resource}.md
+        sed -i '' '/^For party type abbreviations/,$d' docs/resources/${resource}.md
```
enzo-crance-statnett, line 429: macOS uses a different version of `sed` that [turns an optional argument into a required one](https://github.com/lmquang/til/issues/18).

docker-compose.yml
krilor: This change cannot be made in isolation. Needs to be changed in e.g. nginx as well.

krilor (changes requested):
Minor changes like these are OK as long as they are not conflicting with our stuff.
Need to fix the port stuff tho.

krilor (approved):

PR #45, opened 2025-02-14
feat: limit CU-SP updates as SP to valid time within the 2 last weeks (FLEX-463)

enzo-crance-statnett:
With test dependent on today's date, so it does not become outdated.

### DIFF
--- a/db/flex/controllable_unit_service_provider.sql
+++ b/db/flex/controllable_unit_service_provider.sql
@@ -47,3 +47,32 @@ CREATE OR REPLACE TRIGGER controllable_unit_service_provider_event
 AFTER INSERT OR UPDATE OR DELETE ON controllable_unit_service_provider
 FOR EACH ROW
 EXECUTE FUNCTION capture_event('controllable_unit_service_provider');
+
+CREATE OR REPLACE FUNCTION controllable_unit_service_provider_sp_time_limit()
+RETURNS trigger
+SECURITY INVOKER
+LANGUAGE plpgsql
+AS $$
+BEGIN
+    IF (
+        upper(OLD.valid_time_range) IS NOT NULL
+        AND upper(OLD.valid_time_range) < current_timestamp - '2 weeks'::interval
+    ) THEN
+        RAISE sqlstate 'PT400' using
+            message = 'The contract is too old and cannot be modified';
+        RETURN null;
+    END IF;
+
+    IF TG_OP = 'DELETE' THEN
+        RETURN null;
+    ELSE
+        RETURN NEW;
+    END IF;
+END;
+$$;
+
+CREATE OR REPLACE TRIGGER controllable_unit_service_provider_sp_time_limit
+AFTER INSERT OR UPDATE OR DELETE ON controllable_unit_service_provider
+FOR EACH ROW
+WHEN (current_role = 'flex_service_provider')
+EXECUTE FUNCTION controllable_unit_service_provider_sp_time_limit();
--- a/test/api_client_tests/test_cusp.py
+++ b/test/api_client_tests/test_cusp.py
@@ -28,6 +28,7 @@ from flex.api.controllable_unit_service_provider import (
 )
 import pytest
 from typing import cast
+from datetime import date, timedelta
 
 
 @pytest.fixture
@@ -179,11 +180,10 @@ def test_cusp_fiso(data):
 def test_cusp_sp(data):
     (sts, cu_id) = data
 
-    client_fiso = sts.get_client(TestEntity.TEST, "FISO")
     client_sp = sts.get_client(TestEntity.TEST, "SP")
     sp_id = sts.get_userinfo(client_sp)["party_id"]
 
-    # check SP can CRUD the CU-SP relations they are responsible for
+    # check SP can read the CU-SP relations they are responsible for
 
     cusps_sp = list_controllable_unit_service_provider.sync(
         client=client_sp,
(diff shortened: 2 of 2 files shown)


db/flex/controllable_unit_service_provider.sql
```diff
@@ -47,3 +47,32 @@ CREATE OR REPLACE TRIGGER controllable_unit_service_provider_event
 AFTER INSERT OR UPDATE OR DELETE ON controllable_unit_service_provider
 FOR EACH ROW
 EXECUTE FUNCTION capture_event('controllable_unit_service_provider');
+
+CREATE OR REPLACE FUNCTION controllable_unit_service_provider_sp_time_limit()
```
krilor, line 51: Idea: We can make this a generic trigger with number of days as input to stick into the pg_timeline package.

db/flex/controllable_unit_service_provider.sql
```diff
@@ -47,3 +47,32 @@ CREATE OR REPLACE TRIGGER controllable_unit_service_provider_event
...
+    IF TG_OP = 'DELETE' THEN
+        RETURN null;
+    ELSE
+        RETURN NEW;
+    END IF;
+END;
+$$;
+
+CREATE OR REPLACE TRIGGER controllable_unit_service_provider_sp_time_limit
+AFTER INSERT OR UPDATE OR DELETE ON controllable_unit_service_provider
```
krilor, line 75: Not relevant for DELETE

db/flex/controllable_unit_service_provider.sql
```diff
@@ -47,3 +47,32 @@ CREATE OR REPLACE TRIGGER controllable_unit_service_provider_event
...
+SECURITY INVOKER
+LANGUAGE plpgsql
+AS $$
+BEGIN
+    IF (
+        upper(OLD.valid_time_range) IS NOT NULL
+        AND upper(OLD.valid_time_range) < current_timestamp - '2 weeks'::interval
+    ) THEN
+        RAISE sqlstate 'PT400' using
+            message = 'The contract is too old and cannot be modified';
```
krilor, line 62: We need to improve this error message.
It should include "valid time".
Maybe something like "Valid time on contract cannot be set more than 2 weeks back in time"

db/flex/controllable_unit_service_provider.sql
```diff
@@ -47,3 +47,32 @@ CREATE OR REPLACE TRIGGER controllable_unit_service_provider_event
...
 EXECUTE FUNCTION capture_event('controllable_unit_service_provider');
+
+CREATE OR REPLACE FUNCTION controllable_unit_service_provider_sp_time_limit()
+RETURNS trigger
+SECURITY INVOKER
+LANGUAGE plpgsql
+AS $$
+BEGIN
+    IF (
+        upper(OLD.valid_time_range) IS NOT NULL
```
krilor, line 58: Should we check both lower and upper?

db/flex/controllable_unit_service_provider.sql
enzo-crance-statnett: Went for the general solution so the error message also depends on the restricted interval :+1:

db/flex/controllable_unit_service_provider.sql
enzo-crance-statnett: Indeed, if the upper bound is null, the trigger should also complain if the contract started before the limit.

db/pg_timeline/pg_timeline.sql
```diff
@@ -207,3 +207,81 @@ BEGIN
...
+    -- interval allowed back in time
+    tl_freeze_interval text := TG_ARGV[0];
+    -- time field updates
+    tl_time_updates time_update[] := ARRAY[]::time_update[];
+    tl_update time_update;
+BEGIN
+    IF TG_OP = 'INSERT' THEN
+        tl_time_updates := ARRAY[
+            (null, lower(NEW.valid_time_range)),
+            (null, upper(NEW.valid_time_range))
```
krilor, line 240: Can we discuss trying to simplify the implementation here a bit?
This is clever but I would sacrifice a bit of verbosity for a bit of clarity.

Possibilities:

1. Separate checks for insert an update. We only need to check lower bound on insert and it allows us to give better error messages.
2. In the update check, we can possibly split the "update of something that is frozen" and "set valid time to a frozen time" checks to provide better error messages. We might still want to do some looping, but writing out the checks might actually be less LOC overall.

db/pg_timeline/pg_timeline.sql
```diff
@@ -207,3 +207,81 @@ BEGIN
...
+-- freeze a 'valid time' timeline before a certain date :
+--   no date can be inserted or updated before the limit
+CREATE OR REPLACE FUNCTION timeline_freeze()
+RETURNS trigger
+SECURITY INVOKER
+LANGUAGE plpgsql
+AS $$
+DECLARE
+    -- interval allowed back in time
+    tl_freeze_interval text := TG_ARGV[0];
```
krilor, line 232: naming: tl_freeze_after_interval <- could this help with making it more clear what we are using this for?

krilor (commented):
Very nice work on the tests :) Just commenting since we are still missing some docs on this.

db/pg_timeline/pg_timeline.sql
```diff
@@ -207,3 +207,90 @@ BEGIN
...
+                message = 'Cannot create new contract '
+                    || 'more than ' || tl_freeze_after_interval
+                    || ' back in time';
+            RETURN null;
+        END IF;
+    END IF;
+
+    IF TG_OP = 'UPDATE' THEN
+        FOREACH tl_update IN ARRAY
+            ARRAY[
```
krilor, line 251: NIT: One way to get rid of the custom type and ARRAY ARRAY :: casting is to just do a loop over select

```sql
DECLARE
    tl_update record;
BEGIN
...
FOR tl_update IN
  select lower(OLD.valid_time_range) as old, lower(NEW.valid_time_range) as new
  union all
  select upper(OLD.valid_time_range) as old, upper(NEW.valid_time_range) as new
LOOP
...
FOR 
```

Not sure if it is "better", but it doesnt clutter the pg types at least.

docs/time.md
krilor, line 1: Very nice!

krilor (approved):
Nice and clear! 👍

PR #46, opened 2025-02-17
feat(db): midnight alignment (FLEX-462)

enzo-crance-statnett:
- [x] trigger in `pg_timeline`
- [x] SQL unit tests
- [x] enable the trigger on all contract resources and document it
- [x] fix API tests if needed
- [x] only show/input dates in the UI (input components to store Norwegian midnight internally)
- [x] docs


db/pg_timeline/pg_timeline.sql
```diff
@@ -284,3 +284,33 @@ BEGIN
...
+LANGUAGE plpgsql
+AS $$
+DECLARE
+    -- the timestamp to check
+    tl_timestamp timestamptz;
+    -- the same timestamp in Norwegian timezone
+    tl_norwegian_timestamp timestamptz;
+BEGIN
+    FOREACH tl_timestamp IN ARRAY
+        ARRAY[lower(NEW.valid_time_range), upper(NEW.valid_time_range)]
```
krilor, line 300: How does this behave on nulls?

db/pg_timeline/pg_timeline.sql
```diff
@@ -284,3 +284,33 @@ BEGIN
...
+    LOOP
+        tl_norwegian_timestamp := tl_timestamp at time zone 'Europe/Oslo';
+        IF (
+            extract(hour from tl_norwegian_timestamp) != 0
```
krilor, line 304: This does not check for anything less than seconds. We should consider using date_trunc instead

```sql
date_trunc('day', tl_norwegian_timestamp) != tl_norwegian_timestamp
```

enzo-crance-statnett: `second` is actually a microsecond precision value. The [official documentation](https://www.postgresql.org/docs/current/datatype-datetime.html) confirms this, so we are not missing anything with the current code.

Switching to your proposition for conciseness.

krilor: TIL

db/pg_timeline/pg_timeline.sql
enzo-crance-statnett: We get an `IF null` and I do not know how it behaves. I added a check to be sure.

db/pg_timeline/test/20_midnight.sql
```diff
@@ -0,0 +1,99 @@
...
+select plan(7);
+
+-- noqa: disable=all
+
+-- insert midnight values in both UTC and Norwegian timezones : OK
+
+prepare try_insert_abcdef as
+insert into pgtl_midnight (tl_id, value, valid_time_range)
+values
+(1, 'a', tstzrange('2025-01-01 23:00:00+0', '2025-01-05 23:00:00+0')),
```
krilor, line 28: Shouldn't we have a case here with upper end null?
There is also the special infinity value that we should test for, I think.

db/pg_timeline/pg_timeline.sql
```diff
@@ -284,3 +284,31 @@ BEGIN
...
+    -- the timestamp to check
+    tl_timestamp timestamptz;
+    -- the same timestamp in Norwegian timezone
+    tl_norwegian_timestamp timestamptz;
+BEGIN
+    FOREACH tl_timestamp IN ARRAY
+        ARRAY[lower(NEW.valid_time_range), upper(NEW.valid_time_range)]
+    LOOP
+        tl_norwegian_timestamp := tl_timestamp at time zone 'Europe/Oslo';
+        IF tl_norwegian_timestamp IS NOT null
```
krilor, line 303: Seems a bit late for a null check. How does the "null at time zone" above work? Maybe we should we do  the nullcheck early in the loop or remove nulls with array_remove in the FOREACH call ?

db/test_data/test_data.sql
```diff
@@ -512,9 +512,9 @@ BEGIN
       so_id,
       add_check_digit(accounting_point_seq::text),
       ARRAY[
-          (sp_id, '[2024-07-01 09:00:00 CET,2024-08-01 09:00:00 CET)'::tstzrange),
-          (common_sp_id, '[2024-08-01 09:00:00 CET,2024-09-01 09:00:00 CET)'::tstzrange),
-          (sp_id, '[2024-09-01 09:00:00 CET,)'::tstzrange)
+          (sp_id, '[2024-07-01 00:00:00 CEST,2024-08-01 00:00:00 CEST)'::tstzrange),
+          (common_sp_id, '[2024-08-01 00:00:00 CEST,2024-09-01 00:00:00 CEST)'::tstzrange),
```
krilor, line 516: Would it be simpler to just write `Europe/Oslo` over `CEST` / `CET` so we don't have to know or check what is summer and winter time?

db/pg_timeline/test/20_midnight.sql
enzo-crance-statnett: Added both : the unfinished contract (null upper bound), and the "infinite" contract which encodes for a deleted entry.

db/test_data/test_data.sql
enzo-crance-statnett: Indeed ! I had forgotten this `Continent/City` format and I did not know it was supported in the ISO strings.

krilor (approved):
🥳

PR #49, opened 2025-02-19
feat(db): valid time window for CU-SP insert as SP (FLEX-466)

enzo-crance-statnett:


db/pg_timeline/pg_timeline.sql
```diff
@@ -229,9 +229,13 @@ BEGIN
         IF lower(NEW.valid_time_range) < tl_freeze_time
         THEN
             RAISE sqlstate 'PT400' using
-                message = 'Cannot create new contract '
-                    || 'more than ' || tl_freeze_after_interval
-                    || ' back in time';
+                message = 'Cannot create new contract more than '
+                    || tl_freeze_after_interval || ' back in time (latest '
+                    || to_char(
+                           tl_freeze_time at time zone 'Europe/Oslo',
```
krilor, line 235: Do we think that this is the right thing to do? I like the idea, but if we pair this with the midnight alignment, then the error message _might_ be a bit confusing. Client will go "Oh, I'll just use this time then" and hit another error. Thoughts?

krilor: One idea could be to say that all these checks are midnight aligned as well and date_trunc tl_freeze_time from the get-go.

db/pg_timeline/pg_timeline.sql
```diff
@@ -312,3 +326,54 @@ BEGIN
...
+--   a contract can neither be created too short or too long in advance
+--   (i.e., only after a first interval has passed and during a second interval)
+CREATE OR REPLACE FUNCTION timeline_window()
+RETURNS trigger
+SECURITY INVOKER
+LANGUAGE plpgsql
+AS $$
+DECLARE
+    -- using start + duration ensures the end of the window is after its start
+    tl_window_start_interval text := TG_ARGV[0];
```
krilor, line 340: NIT: Should we add an `after_` here as well to align with the freeze one? `tl_window_start_after_interval`
NIT2: `duration_interval` is "smør på flesk" just interval might be enough.

db/pg_timeline/pg_timeline.sql
```diff
@@ -312,3 +326,54 @@ BEGIN
...
+                       tl_window_start at time zone 'Europe/Oslo',
+                       'DD.MM.YYYY kl. HH24:MI'
+                   )
+                || ')';
+        RETURN null;
+    END IF;
+
+    IF lower(NEW.valid_time_range) > tl_window_end THEN
+        RAISE sqlstate 'PT400' using
+            message = 'Cannot create new future contract more than '
```
krilor, line 363: NIT: Remove "future" to align with the other error message?

krilor (commented):
Lets talk about error messages and (possible) midnight alignment.

If we want to go with providing more details/hints then we should use `detail` or `hint` and not put it in a parentesis in the `message`.

db/pg_timeline/pg_timeline.sql
enzo-crance-statnett: > _Client will go "Oh, I'll just use this time then" and hit another error._

Yes, this behaviour was exactly what I was thinking about when saying _"it may become confusing"_. Midnight-aligning the checks is maybe not that costly to implement right from the start.

Regarding the use of `detail` or `hint`, I understand it from an API perspective, but we have to keep in mind that such details won't end up in the error messages in the UI, as [React Admin uses `message`](https://marmelab.com/react-admin/DataProviderWriting.html#error-format). This may be customisable though.

krilor: I know. But our main focus should always be to make the API we actually want. API first.

db/pg_timeline/pg_timeline.sql
```diff
@@ -208,8 +208,38 @@ BEGIN
...
+AS $$ SELECT
+    -- convert to timestamp in Norwegian timezone to get the right day
+    ((t at time zone 'Europe/Oslo')
+    -- ignore the time and switch to text format
+    ::date::text
+    -- add the Norwegian midnight time
+    || ' Europe/Oslo')
+    -- switch back to timestamp with timezone
+    ::timestamptz
+$$;
```
enzo-crance-statnett, line 225: There is an alternative with a local variable, several calls to `extract`, and a call to `make_timestamptz` allowing not to go through a cast to `text`, but I found this more concise and direct.

krilor: Going via text is probably the worst ting to do from a performance point of view. Still fast but probably the worst :P

I don't really understand why we need a function here. Can't we just do:

```sql
date_trunc('day', mytimestamp, 'Europe/Oslo')
```

Ref https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-TRUNC

> When the input value is of type timestamp with time zone, the truncation is performed with respect to a particular time zone; for example, truncation to day produces a value that is midnight in that zone. By default, truncation is done with respect to the current [TimeZone](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-TIMEZONE) setting, but the optional time_zone argument can be provided to specify a different time zone.

db/pg_timeline/pg_timeline.sql
```diff
@@ -208,8 +208,38 @@ BEGIN
...
+
+-- convert a timestamp to an ISO 8601 string
+--   UTC timezone for simplicity (writing Z instead of computing an offset)
+CREATE OR REPLACE FUNCTION to_iso_string(t timestamptz)
+RETURNS text
+SECURITY INVOKER
+LANGUAGE sql
+AS $$ SELECT
+    to_char(t at time zone 'UTC', 'YYYY-MM-DD')
```
krilor, line 234: This function does not use the functionality that postgres actually has.

```
to_char(mytimestamtz at time zone 'UTC', 'YYYY-MM-DD"T"HH24:MI:SS"Z"')
-- or this one that is not javascript compatible but hey its just error message for a human
to_char(mytimestamtz, 'YYYY-MM-DD"T"HH24:MI:SSOF')
```

THere is also this thing that is pretty nice but a bit ugly

https://stackoverflow.com/a/55387470
```
postgres@localhost:flex> select to_json(date_trunc('day', now(), 'Europe/Oslo'))#>>'{}';
+---------------------------+
| ?column?                  |
|---------------------------|
| 2025-02-19T23:00:00+00:00 |
+---------------------------+
```

We should also remember that this is just a friendly error message in the return and not something that is supposed to machine readable, so using the value at whatever timezone is ok.

db/pg_timeline/pg_timeline.sql
enzo-crance-statnett: TIL for the timezone argument.

This allows not only adding UTC midnight but proper Norwegian midnight after the day. I probably discarded this function because of this in the first place : if it adds the wrong time, you need post-processing to replace it and it's annoying.

I will give it a chance.

docs/time.md
```diff
@@ -108,24 +114,28 @@ non-frozen past, so the third update (`5'''`) is invalid, but the last one is.
 ### Timeline window
 
 For some contracts, _insertion_ can be restricted to a fixed _window_ from the
```
krilor, line 116: NIT: We should have written "timelines" not "contracts" in these sections, I think. This  functionality is agnostic of the type of resource. But things like "as an example, contracts ... " could be good tho.

krilor (approved):

PR #51, opened 2025-02-21
feat(ui): allow creating a CU-SP from the CU ID (FLEX-470)

enzo-crance-statnett:
This PR makes it possible to create a CU-SP contract in the portal without having to choose a CU from the readable ones.

> [!IMPORTANT]
> Funnily, this is only a UI update. CUSP-SP001 covers the need on the db side. Tests already work with SP not being able to read the CU as the added check to the test shows.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -576,6 +576,22 @@ export const App = () => (
             />
           </Resource>
         ) : null}
+        {permissions.includes("controllable_unit_service_provider.read") ? (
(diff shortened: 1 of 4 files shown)


enzo-crance-statnett:
New button in toolbar on CU list

Screenshot 2025-02-21 110641

ID input in this special case

Screenshot 2025-02-21 110700

krilor (approved):
Ilikeitalot

PR #52, opened 2025-02-21
docs: clarify that SPGM-SP001 applies to future contracts (FLEX-467)

enzo-crance-statnett:
The feature is already supported, it was just unclear in the rule.


krilor (approved):
👍

PR #53, opened 2025-02-24
feat(db): ensure consistency between CU-SP and SPG membership valid timelines (FLEX-125)

enzo-crance-statnett:
EDIT (after https://github.com/elhub/flex-information-system/pull/53#issuecomment-2681232351)

- [x] SPGM insert/update RLS policy
- [x] CUSP update/delete trigger
- [x] fixed & green API tests


db/flex/service_providing_group_membership_rls.sql
```diff
@@ -54,6 +54,18 @@ WITH CHECK (
...
+    --     (
+    --         SELECT array_agg(cusp.valid_time_range)
+    --         FROM controllable_unit_service_provider AS cusp
+    --             INNER JOIN service_providing_group AS spg
+    --                 ON cusp.service_provider_id = spg.service_provider_id
+    --         WHERE spg.service_provider_id = current_party()
+    --             AND cusp.controllable_unit_id = service_providing_group_membership.controllable_unit_id -- noqa
+    --             AND spg.id = service_providing_group_membership.service_providing_group_id -- noqa
+    --     )
+    -- )
```
enzo-crance-statnett, line 68: This loop function does not exist but it's a glimpse of what the RLS policy could look like (checking that subtracting all the CU-SP valid time ranges overlapping with the new SPGM relation returns an empty time range in the end -- meaning the relation is fully covered).

enzo-crance-statnett:
> [!IMPORTANT]
> I got sidetracked by the generality of timelines, but here we can make an interesting assumption that [we have already made before](https://github.com/elhub/flex-information-system/blob/5d84653d96721fb8e50eabc4aa8618de9892da1e/db/flex/service_providing_group_membership.sql#L53-L55) when designing SPGM in the first place : 
> 
> _We suppose the CU-SP timeline is clean, meaning we do not have several contracts for the same SP following each other._

This means instead of having a SPGM valid time range _covered_ by the CU-SP _timeline_, we have it _included_ in a _single_ CU-SP valid time _range_.

Here are the simplifications this allows :

- **For INSERT/UPDATE on SPGM**

  We can check inclusion instead of coverage, _i.e._, there exists _one_ relevant record of the CU-SP timeline that fully includes the desired SPGM valid time. No need for multiranges.

- **For DELETE of CU-SP**

  Deleting the CU-SP won't leave SPGMs partially covered. The SPGMs included in the deleted range are now 0% covered, so they are to be deleted as well. No risk of aggressive delete.

- **For UPDATE of CU-SP**

  We can switch to using an intersection operator.

  Let's call `M` the SPGM and `C` the CUSP it belongs to. In the assumed normal situation, `M` is fully included in `C`, so `M ∩ C = M`. But when `C` is modified, we can just recompute this intersection and set this to the new valid time of `M`. This will (I hope) cover all the cases we need, even the tricky ones **(*)**. If the intersection is empty, we can delete the SPGM.

  **(*)** If the new CU-SP becomes smaller than the former SPGM, the new SPGM will take its shape and this computation is done in one step instead of computing the diff like I first did.

db/flex/controllable_unit_service_provider.sql
```diff
@@ -71,3 +71,70 @@ BEFORE INSERT ON controllable_unit_service_provider
...
+            AND OLD.valid_time_range @> spgm.valid_time_range
+    LOOP
+        l_new_spgm_valid_time_range :=
+            l_spgm.valid_time_range * NEW.valid_time_range;
+        IF (
+            l_new_spgm_valid_time_range IS NULL
+            OR isempty(l_new_spgm_valid_time_range)
+        ) THEN
+            DELETE FROM service_providing_group_membership
+            WHERE id = l_spgm.id;
```
enzo-crance-statnett, line 123: Soft-delete version :
```sql
UPDATE service_providing_group_membership
SET valid_time_range = tstzrange(lower(valid_time_range), lower(valid_time_range), '[)')
WHERE id = l_spgm.id;
```

db/flex/controllable_unit_service_provider.sql
```diff
@@ -71,3 +71,70 @@ BEFORE INSERT ON controllable_unit_service_provider
...
+    l_spgm record;
+    -- new valid time range for these SPGMs
+    l_new_spgm_valid_time_range tstzrange;
+BEGIN
+    -- delete CUSP -> delete SPGM on that period
+    IF TG_OP = 'DELETE' THEN
+        DELETE FROM service_providing_group_membership
+        WHERE controllable_unit_id = OLD.controllable_unit_id
+            AND service_provider_id = OLD.service_provider_id
+            AND OLD.valid_time_range @> valid_time_range;
```
enzo-crance-statnett, line 95: Soft-delete version :
```sql
UPDATE service_providing_group_membership
SET valid_time_range = tstzrange(lower(valid_time_range), lower(valid_time_range), '[)')
WHERE controllable_unit_id = OLD.controllable_unit_id
    AND service_provider_id = OLD.service_provider_id
    AND OLD.valid_time_range @> valid_time_range;
```

enzo-crance-statnett (commented):
After discussion, the trigger could be weakened in the case of SPGMs becoming 0% covered by CU-SP, so that we do not lose them. Examples in comments.

enzo-crance-statnett:
After discussion, this PR becomes obsolete because we are going for a no-op on CU-SP change + notice solution. Objective is to merge https://github.com/elhub/flex-information-system/pull/54 and then start from there to build the notice feature. Closing.

PR #54, opened 2025-02-25
feat(db): ensure SPGM/CU-SP consistency on SPGM insert/update (FLEX-125)

enzo-crance-statnett:
SPGM-only part of https://github.com/elhub/flex-information-system/pull/53 because it is the part we are sure about for now.

Basically not much to change here, just a better join and an additional constraint on the valid time range, which was missing. Indeed, the former policy was wrong because it authorised the operation as soon as there is one contract, it did not check _when_.

The tests, which were already good for our current needs, were only green by chance, so nothing changes on that side.

### DIFF
--- a/db/flex/service_providing_group_membership.sql
+++ b/db/flex/service_providing_group_membership.sql
@@ -56,7 +56,8 @@ BEGIN
 
     IF lv_match_count = 0 THEN
       RAISE 'The CU must be managed by the SP from % to %',
-        lower(NEW.valid_time_range), upper(NEW.valid_time_range);
+        timeline_timestamptz_to_text(lower(NEW.valid_time_range)),
+        timeline_timestamptz_to_text(upper(NEW.valid_time_range));
     END IF;
 
     RETURN NEW;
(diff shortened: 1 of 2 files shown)


db/flex/service_providing_group_membership.sql
```diff
@@ -56,7 +56,8 @@ BEGIN
 
     IF lv_match_count = 0 THEN
       RAISE 'The CU must be managed by the SP from % to %',
-        lower(NEW.valid_time_range), upper(NEW.valid_time_range);
+        timeline_timestamptz_to_text(lower(NEW.valid_time_range)),
+        timeline_timestamptz_to_text(upper(NEW.valid_time_range));
```
enzo-crance-statnett, line 60: Spotted while confirming tests in the UI. Now that we have a function for readable timestamps in error messages, better use it...

krilor (commented):
I was assuming that this PR would contain some fixes of this TODO?

https://github.com/elhub/flex-information-system/blob/main/db/flex/service_providing_group_membership.sql#L53

krilor (approved):
Nice!

PR #55, opened 2025-02-25
feat(backend): notify end user on CU-SP update (FLEX-469)

enzo-crance-statnett:
This PR adds notifications for end users behind a controllable unit during the valid time period of a CU-SP contract when this contract changes. Both the end users _before and after_ the update are notified.

This change supports the delete operation on CU-SP because it is a soft delete, _i.e._, turned into an update.

### DIFF
--- a/backend/event/models.sql
+++ b/backend/event/models.sql
@@ -57,13 +57,32 @@ AND tstzrange(cusph.recorded_at, cusph.replaced_at, '[)') @> @recorded_at::times
 AND tstzrange(cusph.valid_from, cusph.valid_to, '[)') @> @recorded_at::timestamptz;
 
 -- name: GetControllableUnitServiceProviderUpdateDeleteNotificationRecipients :many
-SELECT unnest(
-    array[cusp.service_provider_id, cu.connecting_system_operator_id]
+SELECT DISTINCT unnest(
+    array_remove(
+        array[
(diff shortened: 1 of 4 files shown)


krilor (changes requested):
We are missing defintions in https://github.com/elhub/flex-information-system/blob/main/docs/resources/controllable_unit_service_provider.md

I also don't really understand what "Both the end users before and after the update are notified." mean.

I'm also conscerned about the soft deleted records. What happens if a record is restored? Isn't the soft deleted records stored as `[null,null)`, meaning they overlap all end users?

enzo-crance-statnett:
> "Both the end users before and after the update are notified."_

```
├── 1 ──┼── 2 ──┼── 3 ──┼── 4 ──┤   end user timeline
  ├────────┤                        former CUSP valid time
                         ├───┤      new CUSP valid time
```
End users `1`, `2` and `4` are notified.

It's a proposal though, I don't know what makes most sense to do. @JanMStrand @kelvestad

enzo-crance-statnett:
> _Isn't the soft deleted records stored as `[null,null)`, meaning they overlap all end users?_

Good point.

We could have a `timeline_overlap` function that does the null-check before calling `&&`, so we are sure this case is handled. Because this issue may already also exist elsewhere in the system.

krilor (approved):

PR #56, opened 2025-02-25
feat(backend): lower log level on primary keepalive replication messages (FLEX-464)

enzo-crance-statnett:
This small change reduces the noise in the backend logs. We can do more or less than that :

⏫ A _harder_ approach would be to also hide the specific log of the event worker acquiring a connection everytime it wants to check the events, but it requires a bit more work because this log is actually useful for other roles in the system.

⏬ A _softer_ approach would be to keep the keepalive logs, but only at the `DEBUG` level, so we can filter this out when checking the logs.

### DIFF
--- a/backend/pgrepl/connection.go
+++ b/backend/pgrepl/connection.go
@@ -169,9 +169,14 @@ func (replConn *Connection) handlePrimaryKeepaliveMessage(
 		if err != nil {
 			return sendStandbyStatusUpdateError(err)
 		}
-		slog.InfoContext(ctx, "received PrimaryKeepaliveMessage and sent status update", "lsn", replConn.lsn.String())
+		slog.DebugContext(
+			ctx, "received PrimaryKeepaliveMessage and sent status update",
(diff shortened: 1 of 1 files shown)


backend/pgrepl/connection.go
```diff
@@ -169,9 +169,6 @@ func (replConn *Connection) handlePrimaryKeepaliveMessage(
 		if err != nil {
 			return sendStandbyStatusUpdateError(err)
 		}
-		slog.InfoContext(ctx, "received PrimaryKeepaliveMessage and sent status update", "lsn", replConn.lsn.String())
```
krilor, line 172: Shouldn't we still have this in "Debug" ? So basically just change this to from InfoContext to DebugContext?

I think it is valuable from a debugging perspective.

FLEX-464 is about the running system, not dev env

enzo-crance-statnett:
Switching to the softer approach.

krilor (approved):

PR #58, opened 2025-02-27
feat: notice resource with SPGM valid time inconsistency (FLEX-125)

enzo-crance-statnett:
This PR adds the `notice` resource we are going to use for inconsistency warnings, among others.

It is made from a _view_ in the `flex` schema to define the notices, which is then mapped to the `api` schema where it gets filtered by role. Associated page in the UI, only list mode allowed for now.

Screenshot 2025-02-27 154332

Could add a button on each line to jump to the resource based on our existing component for that.


krilor:
> Could add a button on each line to jump to the resource based on our existing component for that.

Yep. We should.

openapi/resources.yml
```diff
@@ -1247,3 +1247,49 @@ resources:
...
+    acronym: NOTICE
+    summary: Notice
+    description: >-
+      Notice to users about various issues or actions expected from them.
+    operations:
+      - list
+    generate_views: false
+    properties:
+      id:
+        description: NON-STABLE unique identifier.
```
krilor, line 1261: > NON-STABLE

Should we add a bit of x-detail on this point? Its uppercase and all, but we must stress this point even more and provide a bit of context about it?

krilor (dismissed):

PR #59, opened 2025-02-28
feat(db): time dependent RLS for technical resource (FLEX-459)

enzo-crance-statnett:
Policies based on CU policies. As the CU RLS is more permissive and not time dependent for SO, it is currently the same for TR.


db/api/technical_resource.sql
```diff
@@ -40,3 +28,128 @@ WITH (
...
+                    message = 'Unauthorized: service provider is not responsible for the controllable unit';
+            ELSE
+                RAISE sqlstate 'PT401' using
+                    message = 'Unauthorized';
+            END IF;
+            RETURN null;
+        END IF;
+
+        SELECT * INTO l_new FROM api.technical_resource WHERE id = NEW.id;
+        return l_new;
```
krilor, line 130: NIT: Something is up with the formatting here.
The indentation of ELSIF breaks the reading flow and return should be uppercase. Is the formatter not fixing that?

db/api/controllable_unit.sql
```diff
@@ -253,7 +253,7 @@ BEGIN
             RETURN null;
         END IF;
 
-        RETURN NULL;
+        RETURN OLD;
```
krilor, line 256: Having a delete return the OLD value, is that "correct" behaviour? Doesn't empty response make more sense?

krilor (dismissed):

db/api/controllable_unit.sql
krilor: Nevermind 🙈
https://stackoverflow.com/a/49546159

db/api/technical_resource.sql
enzo-crance-statnett: I checked while fixing the return value, and sqlfluff shows `Passed`, not `Skipped`. So it should be running on these files.

For the `ELSIF` I don't see the issue as it's aligned with `IF` and `END IF`. Maybe the blank lines around ?

We can consider making the linter stricter to fix this later.

PR #62, opened 2025-03-04
fix(backend): fix schemathesis error on Content-Type (FLEX-465)

enzo-crance-statnett:
This PR corrects one of the schemathesis errors about the PostgREST default header for `Content-Type` not being aligned with our specification (`application/json`). This makes schemathesis a bit greener, and keeps API tests green.

It works by using a middleware that "captures" headers and body, and changes the `Content-Type` header to the right value before everything is written and sent over the network.


enzo-crance-statnett (commented):
I still do not fully understand the flow of response writing. There are side effects everywhere and it is hard to be sure I have written all required overrides.

Indeed, a Gin handler has some methods in the context that trigger _writing_ the response directly to a buffer for sending it over the network when they are called. This means once the headers or body are written, they somehow cannot be overwritten, so we have to build a custom response writer that overrides any possible method forcing things to be written too soon to the response buffer, and then revert to the default version of these methods after our middlewares have all run and done their modifications.

I guess this implementation was thought for performance reasons (write directly to the raw buffer instead of building a response object that gets written in the end), but this makes it harder to track what happens when and to think of middlewares as a composition of functions.

enzo-crance-statnett:
Also, I wanted this middleware to be a test to clarify whether the other bug on `Content-Type` comes from outside of Gin, but I really cannot conclude anything with confidence right now.

krilor (commented):
Is it possible to put everything in one handler, to avoid capturing the response twice?

Its probably not a big deal, but depending on the request it might be a bit of data.

Basically a rename like `ErrorMessageMiddleware` to `PostgRESTResponseMiddleware` or something?

krilor (approved):

PR #65, opened 2025-03-05
feat: token out of localstorage (FLEX-475)

krilor:
- [x] openapi spec
- [x] prune npm dependencies (jwt decoding)

### DIFF
--- a/frontend/src/auth/authProvider.tsx
+++ b/frontend/src/auth/authProvider.tsx
@@ -1,7 +1,6 @@
-import { AuthProvider as RAAuthProvider, fetchUtils } from "react-admin";
-import { jwtDecode } from "jwt-decode";
+import { AuthProvider, fetchUtils, getStorage } from "react-admin";
 import permissions from "./permissions.json";
-import { httpClient, authURL } from "../httpConfig";
+import { authURL } from "../httpConfig";
 
 import anonymous_avatar from "./avatars/ANO.png";
 import balance_responsible_party_avatar from "./avatars/BRP.png";
@@ -13,11 +12,6 @@ import market_operator_avatar from "./avatars/MO.png";
 import service_provider_avatar from "./avatars/SP.png";
 import system_operator_avatar from "./avatars/SO.png";
 import third_party_avatar from "./avatars/TP.png";
-
-export enum GrantType {
-  ClientCredentials,
-  TokenExchange,
-}
 
 const toDataURL = (url: string) =>
   fetch(url)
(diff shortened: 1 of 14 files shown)


frontend/src/auth/authProvider.tsx
```diff
@@ -46,174 +40,154 @@ const roleAvatars: any = {
...
 export function authProvider(): AuthProvider {
   const getIdentity = async () => {
-    if (!localStorage.getItem("token")) return Promise.reject();
+    const sessionInfoString = getStorage().getItem(sessionInfoKey);
+    if (!sessionInfoString) return Promise.reject();
 
-    const { json } = await httpClient(`${authURL}/userinfo`).catch(() => {
-      localStorage.removeItem("token");
-      return Promise.reject();
-    });
```
enzo-crance-statnett, line 60: Our former way of establishing identity was quite conservative : the `getIdentity` method always did a network call, so we were 100% sure the identity we got was always fresh, even though we used the network more often than necessary.

I'm not saying that the currently proposed change is wrong, but I am still not confident about how often and in which contexts React-Admin calls this method internally, so we need to make sure it is right to go back to local-only checks.

Putting this comment as a pointer for later in the review process / discussions.

enzo-crance-statnett (changes requested):
I confirm some errors show on my machine as well. IMHO this is not ready because some of the current errors should not make it into the app.

I got several weird delays around the login flow with Authelia, especially on the way back, and even an error page stopping the auth flow completely :
```json
{
  "code":"HTTP400",
  "message":"Bad Request",
  "details":"missing login cookie",
  "hint":"http: named cookie not present"
}
```
I have a hard time trying to reproduce consistently or track where this one comes from.

I also got the following one :
```json
{
  "code":"HTTP500",
  "message":"Internal Server Error",
  "details":"could not begin tx in callback handler",
  "hint":"failed to acquire connection for role flex_entity: context deadline exceeded"
}
```
but this one came from a panic in the backend that can be traced back to an error conditional that should return but did not. Fixed in a commit I took the liberty to add to this branch (it's very small).

The third one comes once in the app and it's the one we talked about. Most auth features seem to work in the UI, but I almost always get `Authentication failed` after the first click on the assume party button.

The procedure behind the button manages to store the new cookie, but it does not update the information in the web app (we are not redirected and the identity in the corner remains the same). If we reload the page, we get only one possible party, the one we already assumed. I think this confirms the cookie is well updated. As we are not redirected, maybe it means the first login call actually fails even though it manages to set the right cookie ?

enzo-crance-statnett (approved):

PR #66, opened 2025-03-05
feat(db): implement other cases of notices (FLEX-479)

enzo-crance-statnett:
This PR adds the other kinds of notices we talked about. Tested in the UI.

### DIFF
--- a/db/flex/notice.sql
+++ b/db/flex/notice.sql
@@ -1,5 +1,56 @@
 CREATE OR REPLACE VIEW notice AS (
+    -- CU grid node ID missing
+    SELECT
+        row_number() OVER () AS id,
+        apt.system_operator_id AS party_id,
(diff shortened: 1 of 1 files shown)


db/flex/notice.sql
```diff
@@ -1,5 +1,56 @@
...
+        null AS data -- noqa
+    FROM controllable_unit AS cu
+        INNER JOIN accounting_point AS apt
+            ON cu.accounting_point_id = apt.business_id
+    WHERE cu.grid_node_id IS null
+
+    -- CU grid validation status pending
+    UNION ALL
+    SELECT
+        row_number() OVER () AS id,
```
krilor, line 17: Should we make a subselect and make row_number() as part of the outer stuff? I'm thinking that we will be getting duplicate numbers with the current approach.

(yes its not a big deal since they are super-unstable)

krilor (dismissed):
🥳

PR #68, opened 2025-03-06
feat(backend): PostgREST middleware as stdlib response modifier (FLEX-483)

enzo-crance-statnett:
This PR turns the response-rewriting mechanism that runs when coming back from PostgREST into a standard library `ReverseProxy` response modifier.

It means it is no longer a middleware but a part of the PostgREST handler, and we no longer rely on a custom body capture mechanism.

### DIFF
--- a/backend/data/data.go
+++ b/backend/data/data.go
@@ -5,11 +5,13 @@ import (
 	"encoding/json"
 	"flex/pgpool"
 	"fmt"
+	"io"
 	"log/slog"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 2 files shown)


backend/data/data.go
```diff
@@ -133,68 +110,70 @@ type errorMessage struct {
...
+		"data API failure",
+		"code", errorBody.Code,
+		"message", errorBody.Message,
+		"detail", errorBody.Detail,
+		"hint", errorBody.Hint,
+	)
+	return nil
+}
+
+func writeErrorResponse(rsp *http.Response, msg errorMessage) {
```
krilor, line 174: NIT: Missing doc comment. Its really nice to have those when reading the code. E.g. when hovering over symbols. We should have it on all methods/functions IMHO.

krilor (dismissed):
👍

PR #70, opened 2025-03-06
feat: controllable unit lookup (FLEX-478)

enzo-crance-statnett:
This PR adds the CU lookup feature.

The definition of handlers related to the data API is done in the data module, and this module does no longer depend on Gin.


backend/data/data.go
```diff
@@ -13,19 +13,19 @@ import (
...
 type API struct {
+	PathPrefix   string
 	postgRESTURL *url.URL
 	db           *pgpool.Pool
 	ctxKey       string
 }
 
 // NewAPI creates a new data.API instance.
 func NewAPI(
+	pathPrefix string,
```
krilor, line 28: Should we rather strip the prefix in the calling mux using https://pkg.go.dev/net/http#StripPrefix ?

backend/data/data.go
```diff
@@ -36,31 +36,86 @@ func NewAPI(
...
 }
 
-// PostgRESTHandler forwards the request to the PostgREST API.
-func (data *API) PostgRESTHandler(ctx *gin.Context) {
+// ServeHTTP regroups all data API handlers into one, while implementing the
+// http.Handler interface.
+func (data *API) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	req.URL.Path = strings.TrimPrefix(req.URL.Path, data.PathPrefix)
+
+	if req.URL.Path == "/controllable_unit" &&
```
krilor, line 51: Wouldn't it make more sense to use a mux (router) here?

backend/data/data.go
```diff
@@ -36,31 +36,86 @@ func NewAPI(
...
+		})
+		return
+	}
+
+	http.Error(w, "Not implemented: CU lookup", http.StatusNotImplemented)
+}
+
+// writeErrorToResponseWriter writes an error message as JSON in the response
+// writer.
+func writeErrorToResponseWriter(
```
krilor, line 87: NIT: consider following the signature of https://pkg.go.dev/net/http#Error ?

krilor (commented):
Good start!

enzo-crance-statnett:
Added UI page to prepare the CU lookup.

It's only reachable from the CU list page, it's a simple form with the values that end up in the body of the CU lookup request. The submit click makes the API call, and will redirect to a dedicated show page or display an error, based on the response. The error case is handled, the valid case not yet because it's not working on the DB side.

A few screenshots :

- CU list page (see toolbar)

Screenshot 2025-03-07 170247

- CU lookup input page

Screenshot 2025-03-07 170312

- error message extracted from the JSON, like with the PostgREST calls

Screenshot 2025-03-07 170618

backend/flex.go
```diff
@@ -222,7 +230,7 @@ func Run(ctx context.Context, lookupenv func(string) (string, bool)) error { //n
 	authAPI := auth.NewAPI(authAPIBaseURL, dbPool, jwtSecret, oidcProvider, requestDetailsContextKey)
 
 	slog.DebugContext(ctx, "Creating data API")
-	dataAPI, err := data.NewAPI(postgRESTUpstream, dbPool, requestDetailsContextKey)
+	dataAPI, err := data.NewAPI("/api/v0", postgRESTUpstream, dbPool, requestDetailsContextKey)
```
krilor, line 233: Why do we want the service to know about the prefix?

db/api/controllable_unit_lookup.sql
```diff
@@ -0,0 +1,40 @@
...
+FROM flex.controllable_unit AS cu
+    INNER JOIN flex.accounting_point AS ap
+        ON cu.accounting_point_id = ap.business_id
+    INNER JOIN flex.accounting_point_end_user AS apeu
+        ON ap.id = apeu.accounting_point_id
+WHERE apeu.valid_time_range @> current_timestamp
+    AND apeu.end_user_id = l_end_user_id
+    AND (l_business_id IS null OR cu.business_id = l_business_id::uuid)
+    AND (l_accounting_point_id IS null OR cu.accounting_point_id = l_accounting_point_id)
+LIMIT 1;
```
krilor, line 39: Why is there a LIMIT 1 here?

enzo-crance-statnett:
Added first simple version of the result page, showing each CU in a Card component. Technical resources printed as JSON, we can improve this later.

Screenshot 2025-03-10 133627

Also fixed the way undefined/null values are handled in the UI form, and made sure the JSON format for technical resources aligns with the API spec.

db/api/controllable_unit_lookup.sql
enzo-crance-statnett: Silly mistake, of course it's a list... fixed.

enzo-crance-statnett:
The function is now defined in the `flex` schema and mirrored in the `api` schema, in order to avoid auth issues. I found this cleaner than the double load of grants.

enzo-crance-statnett:
Screenshot 2025-03-10 155500

backend/flex.go
enzo-crance-statnett: As agreed, we can just return a handler instead of an object coming from the data package. Fixed.

backend/data/models/models.custom.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"context"
+	"encoding/json"
+)
+
+// MarshallableTechnicalResource takes the fields of technical resource that we
+// return from the database in the controllable unit lookup operation, adding
+// the right annotations so that the type can be marshalled to JSON.
+type MarshallableTechnicalResource struct {
```
krilor, line 11: There is something in the naming and comments here that is "off". I wonder if the underlying problem is that we are defining all of this in one "models" package.

It is reasonable and advisable to have different structs for storage (db) and presentation (api), but putting it all in one package is causing us to use names like `ControllableUnitLookup`/`ControllableUnitLookupInternal` and `MarshallableTechnicalResource`/`TechnicalResource`.

Should we consider separating the different concerns into separate submodules?

backend/flex.go
```diff
@@ -40,6 +40,14 @@ func WrapHandlerFunc(f http.HandlerFunc) gin.HandlerFunc {
 	}
 }
 
+// WrapHandler adapts WrapHandlerFunc so that it works on general http.Handler
+// instead of only functions.
```
krilor, line 44: NIT: This doc comment should maybe just explain what the function is rather than pointing to the definition of another function?

db/flex/controllable_unit_lookup.sql
```diff
@@ -0,0 +1,50 @@
...
+FROM flex.controllable_unit AS cu
+    INNER JOIN flex.accounting_point AS ap
+        ON cu.accounting_point_id = ap.business_id
+    INNER JOIN flex.accounting_point_end_user AS apeu
+        ON ap.id = apeu.accounting_point_id
+WHERE apeu.valid_time_range @> current_timestamp
+    AND apeu.end_user_id = l_end_user_id
+    -- one of these fields must not be empty
+    AND (
+        coalesce(l_business_id, '') <> ''
```
krilor, line 39: Are we putting a bit to much business logic here in this function?
Does it make more sense to check this in the app and return som helpful error messages to the user?
Consider using [the ](https://github.com/elhub/flex-information-system/blob/718a292efb9afadcf6d5d898aca7ab70216526cd/backend/internal/validate/validate.go)

By the time the request hits the database, we should expect that the rules are obeyed?

If that means we don't need to care about "emtpy", just "null", then the conditions simplify to

```
cu.accounting_point_id = coalesce(l_accounting_point_id, cu.accounting_point_id)
```

test/inspect/culookup.hurl
krilor, line 1: The file name here is less descriptive than the other file names in this directory. Look at the others and change accordingly.

Also, the hurl stuff here is `inspect` stuff, not tests. We test with python and put it as part of the test suite. The intended use here is just to look at responses - inspect them.

We should move tests to pytest

test/inspect/culookup.hurl
```diff
@@ -0,0 +1,132 @@
...
+[FormParams]
+grant_type: client_credentials
+client_id: test.suite@flex.test
+client_secret: 87h87hijhulO
+HTTP 200
+
+[Captures]
+access_token: jsonpath "$.access_token"
+
+POST {{ auth_url }}/assume
```
krilor, line 14: We should use token exchange when using tokens. Assume is just frontend bits. Look at the token exchange or userinfo inpects for reference.

test/inspect/culookup.hurl
enzo-crance-statnett: I used Hurl because the CU lookup operation is a non-standard one, and inspecting was useful in such a context. I just kept thinking in that same file for building the test.

However, although the OpenAPI client does not contain functionality for it yet, the current test may be ported to Python easily once the CU lookup endpoint is included in the spec.

db/flex/controllable_unit_lookup.sql
enzo-crance-statnett: The frontend already filters null/undefined values and only sends defined strings (cf `lookup` function in the CU lookup input page). This means that if both kinds of CU identifier are empty, we already get an error message in the UI, so we don't need to add anything else.

On the Go side, the [`FormValue`](https://pkg.go.dev/net/http#Request.FormValue) method used to extract values from the URL-encoded body returns empty strings for things that are not found in the form, and the `string` type is not nullable. This means that until we enter the SQL world, all these empty cases will be represented as an empty string. I added a `nullif` check so that we turn back to a null value if the string is empty.

Therefore, in the end, in the DB function, all empty values are represented as `null`, including empty strings (if a user forges an API call sending an empty string, we ignore it -- if we get 2 empty strings, we consider the call wrong and the user will get the error message about missing IDs).

backend/data/data.go
```diff
@@ -68,6 +138,29 @@ func (data *api) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		slog.ErrorContext(ctx, "no request details in context", "error", err)
+		writeErrorToResponseWriter(w, http.StatusInternalServerError, errorMessage{ //nolint:exhaustruct
+			Message: "request is not authenticated",
+		})
+		return
+	}
+
+	allowedRoles := []string{
+		"flex_service_provider",
```
krilor, line 153: We should revisit https://docs.postgrest.org/en/v12/explanations/db_authz.html#functions

frontend/src/controllable_unit/ControllableUnitList.tsx
```diff
@@ -39,6 +40,13 @@ const CULookupButton = () => (
 export const ControllableUnitList = () => {
   const { permissions } = usePermissions();
 
+  const { data: identity, isLoading: identityLoading } = useGetIdentity();
+  if (identityLoading) return <>Loading...</>;
+
+  const isSPOrFISO =
```
krilor, line 46: Any reason why this permission should not be part of the permission list as "controllable_unit.lookup" or (if we want to do it as as separate resource) "controllable_unit_lookup" ?

I just want to avoid special handling wherever we can.

backend/data/data.go
```diff
@@ -304,3 +395,9 @@ func writeErrorToResponse(rsp *http.Response, msg errorMessage) {
 	rsp.ContentLength = int64(len(body))
 	rsp.Header.Set("Content-Length", strconv.Itoa(len(body)))
 }
+
+func (data *api) notFoundHandler(w http.ResponseWriter, req *http.Request) {
```
krilor, line 399: missing doc comment

backend/data/data.go
```diff
@@ -54,8 +61,71 @@ func NewAPIHandler(
...
+	mux.HandleFunc("/identity/", data.postgRESTHandler)
+	mux.HandleFunc("/notice", data.postgRESTHandler)
+	mux.HandleFunc("/notice/", data.postgRESTHandler)
+	mux.HandleFunc("/notification", data.postgRESTHandler)
+	mux.HandleFunc("/notification/", data.postgRESTHandler)
+	mux.HandleFunc("/party", data.postgRESTHandler)
+	mux.HandleFunc("/party/", data.postgRESTHandler)
+	mux.HandleFunc("/party_history", data.postgRESTHandler)
+	mux.HandleFunc("/party_history/", data.postgRESTHandler)
+	mux.HandleFunc("/party_membership", data.postgRESTHandler)
```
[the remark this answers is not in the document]
krilor: https://github.com/elhub/flex-information-system/pull/79 ?

PR #73, opened 2025-03-09
docs: more details around modular monolith structure (FLEX-482)

krilor:
Starting point for discussion on what we want to do wrt db/backend structure


krilor:
This has just been sitting here. Will not be merged, but we might want to go back to it and see at some point

PR #76, opened 2025-03-10
feat: add liquibase locally (FLEX-494)

krilor:


db/changelog.sql
enzo-crance-statnett, line 1: Should this be moved to the `test` directory ?

enzo-crance-statnett (approved):

PR #80, opened 2025-03-12
refactor(backend): data module as isolated single http.Handler (FLEX-478)

enzo-crance-statnett:
Tried to isolate the refactoring part of https://github.com/elhub/flex-information-system/pull/70 to get it in first.


backend/data/data.go
```diff
@@ -13,54 +13,140 @@ import (
...
+		mux:          mux,
 		ctxKey:       ctxKey,
-	}, nil
+	}
+
+	// all other requests are forwarded to PostgREST
+	mux.HandleFunc("/accounting_point", data.postgRESTHandler)
+	mux.HandleFunc("/accounting_point/", data.postgRESTHandler)
+	mux.HandleFunc("/controllable_unit", data.postgRESTHandler)
+	mux.HandleFunc("/controllable_unit/", data.postgRESTHandler)
```
enzo-crance-statnett, line 54: According to the [official docs](https://pkg.go.dev/net/http#hdr-Trailing_slash_redirection-ServeMux), the endpoint with ending slash is necessary to act as a wildcard pattern for the rest of the URL in the mux. However, if we only have this one, the tests fail on some cases.

Something like https://github.com/elhub/flex-information-system/pull/79 could help. When it gets merged, we can then simplify this.

krilor: Merged. Please give it a go.

krilor (dismissed):

backend/data/data.go
enzo-crance-statnett: Turns out the mux does a _redirect_ and not just a silent URL _rewrite_ in the case of a _missing_ trailing slash. After discussion, going for an explicit wildcard solution for now.

Paths can become fully explicit and possibly replace the PostgREST URL rewrite mechanism in the future.

enzo-crance-statnett:
> [!NOTE]
> The strip middleware becomes useless for now after our investigations. We can choose to delete it, or keep the file structure for later (possibly moving middlewares there as well like the wrappers for instance).

krilor (approved):

PR #84, opened 2025-03-12
feat: controllable unit lookup (FLEX-478)

enzo-crance-statnett:
This PR cleans/replaces https://github.com/elhub/flex-information-system/pull/70 and introduces the controllable unit lookup feature.

API tests and schemathesis tests are green, but I get authentication problems both in the portal and in the auth tests. I would rather we fix them before this PR is merged because we also need to check the UI links and behaviour are correct, even though it does not change much from the previous PR.


enzo-crance-statnett:
The second remaining point mentioned in the previous PR is authorisation. Indeed, we do manual role filtering and we could have something a bit stronger for that. I haven't thought about it yet.

As the permissions file is generated from the CSV, we could have a `L` for lookup ? It still feels a bit custom though.

enzo-crance-statnett:
Added notifications :

- `lookup` verb for the event type.

- 1 notification per CU in the lookup result.

  I thought about doing 1 per lookup call because it is the same end user receiving all the notifications for a given call, but then we would not be able to fill the `source` field of the event correctly (it expects the ID of a resource that was checked, and there we would have several).

- event data contains the name of the requesting party, because it is the important piece of information we want to give to the end user : _which_ CU was checked (event `source`) and _who_ checked it (event `data`).

backend/data/data.go
```diff
@@ -119,6 +125,119 @@ func (data *api) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+
+	if !slices.Contains(allowedRoles, rd.Role()) {
+		writeErrorToResponseWriter(w, http.StatusUnauthorized, errorMessage{ //nolint:exhaustruct
+			Message: "user cannot perform this operation",
+		})
+		return
+	}
+
+	var cuLookupRequestBody controllableUnitLookupRequest
+	if err = json.NewDecoder(req.Body).Decode(&cuLookupRequestBody); err != nil {
```
krilor, line 156: We must use POST. We should _never_ use GET with body.

Ref: https://stackoverflow.com/questions/978061/http-get-with-request-body

krilor (changes requested):

krilor:
> _(comment from the other PR)_
> 
> > We need to figure out how to structure the relationship between the auth module and the db module wrt this, but passing the context key like we do now seems off.
> 
> I agree, I don't see a reason for us to get the auth details from another key than the one we are passing everywhere. It means we could just remove the parameter and make it internal to the auth module.
> 
> However, as this PR does not actually introduce a change from `main` in that matter (it does not touch auth), I propose to improve this later.

I was thinking it could make sense to do as a separate PR, but not that we need to "wait for merge".

krilor:
> As the permissions file is generated from the CSV, we could have a `L` for lookup ? It still feels a bit custom though.

I think there is value in having consistent pattern for checking authorization on the frontend. L on resource level in the permissions.csv is ok. It does not flow into docs either way.

Isn't it better hard-coding adding "controllable_unit.lookup" to SP and FISO in the permission.json generation ?

db/flex/controllable_unit_lookup.sql
```diff
@@ -0,0 +1,75 @@
...
+    end_user_id bigint,
+    technical_resources jsonb
+)
+SECURITY DEFINER
+LANGUAGE plpgsql
+AS $$
+DECLARE
+    l_event_data jsonb;
+BEGIN
+    SELECT ('{"requesting_party_name": "' || p.name || '"}')::jsonb
```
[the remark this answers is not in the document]
enzo-crance-statnett: Indeed, this can replace the `data` field of the event.

I think this choice of not exposing `recorded_by` dates back to when we did not have a mechanism to show identities nicely in the UI yet. The integer did not mean much then. Now it would make more sense to expose it.

backend/data/data.go
enzo-crance-statnett: The call on the frontend even errored because of this. TIL

db/api/controllable_unit_lookup.sql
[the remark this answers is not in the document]
enzo-crance-statnett: In our previous database architecture, for isolation reasons, we had decided that the `flex` schema would not be accessed from the outside. For this reason, the base schema used by `sqlc` is generated from the resource YAML file, which is aligned with the `api` schema. The database pool used by the data module also has its search path set to `api`. Forwarding the call from `api` to `flex` is a way to keep consistent with the existing architecture.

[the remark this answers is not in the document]
enzo-crance-statnett: I wonder if there's a way to put `flex` in the search path though, so that if a table or function does not exist in the `api` schema, then we pick the `flex` schema. This would spare us the duplication in this PR, without breaking the rest.

backend/data/data.go
```diff
@@ -119,6 +125,131 @@ func (data *api) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+	}
+
+	slog.InfoContext(
+		ctx, "will lookup controllable unit",
+		"end_user_business_id", endUserBusinessID,
+		"accounting_point_id", accountingPointID,
+		"controllable_unit_business_id", controllableUnitBusinessID,
+	)
+
+	conn, err := data.db.Acquire(ctx)
```
krilor, line 200: Unless we want to have multiple transactions on the same connections, just using pool.Begin directly makes sense. It does an aquire and release behind the scenes. Ref https://github.com/elhub/flex-information-system/blob/main/backend/pgpool/pool.go#L103

backend/data/models.go
```diff
@@ -0,0 +1,70 @@
...
+	BusinessID         string              `json:"business_id"`
+	Name               string              `json:"name"`
+	AccountingPointID  string              `json:"accounting_point_id"`
+	EndUserID          int                 `json:"end_user_id"`
+	TechnicalResources []technicalResource `json:"technical_resources"`
+}
+
+// MarshalControllableUnitLookupResult serialises the result of the database
+// controllable unit lookup operation into JSON.
+func MarshalControllableUnitLookupResult(
```
krilor, line 37: This function mixes a few concerns

1. Transforming the storage model into the api model
2. JSON marshalling

This should be split (or, rather, JSON marshalling should maybe not be part of it?

backend/event/models.sql
```diff
@@ -220,3 +220,15 @@ AND sppach.visibility = 'any_party';
 INSERT INTO notification (event_id, party_id)
 VALUES (@event_id, @party_id)
 ON CONFLICT DO NOTHING;
+
+-- name: GetControllableUnitLookupNotificationRecipients :many
+SELECT apeu.end_user_id::bigint
+FROM controllable_unit AS cu
+INNER JOIN accounting_point AS ap ON ap.business_id = cu.accounting_point_id
+LEFT JOIN accounting_point_end_user AS apeu ON apeu.accounting_point_id = ap.id
```
krilor, line 228: Does a left join here really make sense?

db/flex/controllable_unit_lookup.sql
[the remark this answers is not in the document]
krilor: So should we remove it as part of this PR?

openapi/openapi-api-base.yml
```diff
@@ -45,6 +47,106 @@ components:
...
+      type: object
+      properties:
+        end_user_business_id:
+          description: >-
+            The business ID of the current end user on the controllable unit(s)
+            to lookup.
+          format: text
+          type: string
+          # yamllint disable-line rule:line-length
+          pattern: '^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$'
```
krilor, line 62: This pattern is incorrect. We are talking org number or id-number - entity

backend/data/data.go
```diff
@@ -119,6 +125,131 @@ func (data *api) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+	if err != nil {
+		slog.ErrorContext(ctx, "CU lookup query failed", "error", err)
+	}
+	if err != nil || len(cuLookup) == 0 {
+		writeErrorToResponseWriter(w, http.StatusNotFound, errorMessage{ //nolint:exhaustruct
+			Message: "controllable unit not found",
+		})
+		return
+	}
+	if err = tx.Commit(ctx); err != nil {
```
krilor, line 234: shouldn't we just defer and ignore error-checking? There is basically nothing to commit.

krilor (changes requested):

openapi/openapi-api-base.yml
enzo-crance-statnett: This is a pattern for UUIDs, which is the _only_ format we accept for the business ID of _end user parties_, as specified in [PTY-IFV001](https://flex-test.elhub.no/docs/resources/party/#:~:text=Status-,PTY%2DIFV001,-business_id_type%20is%20UUID).

Are you suggesting to use an _entity_ business ID instead of a _party_ business ID ? We would then have to add an extra join with `party_membership` in the query to bridge the gap, because contracts like _accounting point end user_ use _party_ IDs.

I directly assumed party IDs because they are the natural choice with the way we model our resources, but maybe entity makes more sense for this operation.

backend/data/data.go
enzo-crance-statnett: There are events to commit now.

backend/event/models.sql
enzo-crance-statnett: Indeed, if there is no end user on the CU at the time of the event, the row should _disappear_ instead of returning `null`. We have that for other events as well, it's worth checking.

db/flex/controllable_unit_lookup.sql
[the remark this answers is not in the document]
enzo-crance-statnett: After discussion : from an API perspective, knowing which party did the action is the interesting piece of information to return. Returning the identity would work as well, but it would require an extra API call to get the party name, which we can spare users by including that in the event data. Closing.

db/api/controllable_unit_lookup.sql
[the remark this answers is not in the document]
enzo-crance-statnett: After discussion : as we now have everything accessible under the same defining role in the database, the `flex` schema is no longer a necessary step for everything. The CU lookup function can very well live in the `api` schema without needing a `flex` counterpart. Closing.

backend/event/models.sql
enzo-crance-statnett: Fixed and tested for CU lookup. In the other cases, we notify other users at the same time, so we don't want the row to disappear if an end user is not found. We also explicitly filter null values so it's safe to leave as is.

openapi/openapi-api-base.yml
enzo-crance-statnett: After discussion : the UUID in the party resource used for end users has no business meaning, it is mainly there because this field is useful on other party types and we don't want it to be null, while keeping a format that ensures unicity. For CU lookup, we want to use the business ID on the _entity_ resource because it is the one that makes sense in the market. As we have already added the owning `entity_id` to the party resource, we can just use this to link end user party ID to end user entity business ID, instead of going through `party_membership`.

backend/data/data.go
krilor: Ah good point. Shouldn't we return a HTTP 500 if we are not able to commit the events then?

krilor (dismissed):
Now we're talking!

krilor (approved):

PR #86, opened 2025-03-13
docs(process): recovation and termination

kelvestad:
Detailed Revocation and Termination process.


docs/diagrams/revocation.plantuml
```diff
@@ -0,0 +1,36 @@
...
+actor "Service Provider (SP)" as SP
+participant "Flexibility Information System (FIS)" as FIS
+
+== Revocation and Termination ==
+
+SP -> FIS: Register end date
+activate SP
+note left SP
+Data:
+- Service provider
```
krilor, line 21: Techinally, the SP will update the contract resourse controllable_unit_service_provider (CUSP) - https://github.com/elhub/flex-information-system/blob/main/docs/resources/controllable_unit_service_provider.md

In practical terms this means a http request like

```
POST /controllable_unit_service_provider/567
{ "valid_to" : "<some midnight aligned time>" }
```

.. where 567 is the id of the contract.

So technically all the SP is "sending" is the valid to, they just have to put it in the correct place.
```

File: docs/processes/revocation.md [hunk 2]
```diff
@@ -0,0 +1,26 @@
...
+- Implicit contract termination (e.g., move-out) triggered by the metering point
+  administrator.
+- Service contract termination by service provider.
+
+## Prerequisites
+
+- The service provider acts on behalf of the end user in all scenarios.
+- In the event that FIS receives information about implicit contract
+  termination, the end user's access to the affected controllable unit(s) will
+  be revoked.
```
krilor, line 16:
```suggestion
- End users access to the the controllable unit is given based
  on continiously updated enduser information from Elhub
- A controllable unit service provider contract is decoupled from the end user
  and can span multiple endusers from Elhub
```

Not sure tho..

docs/diagrams/revocation.plantuml
```diff
@@ -0,0 +1,56 @@
...
+
+|||
+    
+alt "Revocation by end user"
+    |||
+    EU -> SP: Revoke contract
+    |||
+else "Implicit contract termination"
+    |||
+    MPA -> FIS: Implicit contract termination
```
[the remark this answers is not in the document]
JanMStrand: We have used NCDR wording, termination and change is maybe not the same.

docs/diagrams/revocation.plantuml
```diff
@@ -0,0 +1,56 @@
...
+    |||
+end
+    |||
+
+== Revocation and Termination ==
+
+SP -> FIS: Update existing contract
+activate SP
+note left SP
+Data:
```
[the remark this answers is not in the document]
JanMStrand: Agree.

docs/diagrams/revocation.plantuml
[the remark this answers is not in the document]
kelvestad: I dont think we should call it "Change in enduser". These changes in Elhub is often, but not always, an actual change in enduser. Could also just be termination of active end user, without any new.

docs/processes/revocation.md
kelvestad: I like this one, as it describes the context on end user access: "End users access to the the controllable unit is given based
  on continiously updated enduser information from Elhub".

But Im not sure if I understand what we are saying in the second suggestion. I think we need to discuss this point more in detail together, because Im also not sure how this will actually work.

JanMStrand: In my head this is an ex.: End user moves out, looses the ownership of CU and by extention the CUSP contract must end

Then - how do we write this in the docs?
A controllable unit service provider contract is decoupled from the controllable unit triggered by update from Elhub --> the update can initiate changes on multiple CUSP contracts.

krilor: I agree to that example. But it is also maybe the simplest case/example. I'm thinking more about changes back in time, any type of adjustment/error correction to contracts in Elhub.

+1 for workshopping this next week

docs/diagrams/revocation.plantuml
krilor: My point in this comment is that the current note sais

```
    Data:
    - Service provider
    - Controllable unit
    - End date
```
While technically all you can update on a contract is the valid from/to

Should we change this?

docs/processes/revocation.md
```diff
@@ -0,0 +1,41 @@
+# Service Provider Contract Revocation and Termination
+
+This process is used to terminate the service provider contract for the
+controllable unit. The process can be triggered by one of the following:
+
+- Service contract revocation by end user.
+- Implicit contract termination triggered by the metering point
```
krilor, line 7: Nitpicking, but isn't it triggered by "changed end user information in Elhub (Metering Point Administrator)"

docs/processes/revocation.md
```diff
@@ -0,0 +1,41 @@
...
+  administrator (Elhub).  
+- Service contract termination by service provider.
+
+Updated end user information from Elhub can occur for several reasons, such as:
+
+- Move in of a new end user, which automatically terminates the current end user.
+- Move out of the current end user.
+- Deactivation of an acounting point, which results in the termination of the
+  current end user.
+- Corrections to move ins and move outs, which trigger updates on end users.
```
krilor, line 17: Should we highlight that this is also _back in time_ / past contracts?

krilor (approved):

PR #88, opened 2025-03-13
Added context to notice resource

kelvestad:


krilor:
@kelvestad hint: https://marketplace.visualstudio.com/items?itemName=darkriszty.markdown-table-prettify ;)

docs/resources/notice.md
```diff
@@ -2,6 +2,21 @@
...
+|---------------------------|-----------------------------------------------|-------------------|------------------------------------|
+| CU                        | Grid node ID missing                          | SO                | Update CU with grid node           |
+| CU                        | Grid validation status pending                | SO                | Update grid validation status      |
+| CU                        | Grid validation status incomplete information | SP                | Update missing information         |
+| SP product application    | Status requested                              | SO                | Update application status          |
+| SPG membership            | Inconsistency: SPG contains expired CU(s)     | SP                | Validate and update SPG membership |
+| SPG product application   | Status requested                              | SO                | Update application status          |
+| SPG grid prequalification | Status requested                              | SO                | Update prequalification status     |
+
```
krilor, line 19: Here is a list of questions I wonder about:

1. Should we have the notice type (i.e. "no.elhub.flex.controllable_unit.grid_node_id.missing" ) as part of the overview? Maybe instead of the "resource" column?
2. In responsible party, should we add details where relevant - e.g. "Connecting SO" instead of just "SO" for grid_node_id 
missing ?
3. Mulitple texts in the "Action required" states "Update X status". Is that what is really the requred action? Isn't it more along the lines of "Do the thing and then update the status when you are done."?

PR #92, opened 2025-03-17
feat(ui): tooltips for CU lookup (FLEX-509)

enzo-crance-statnett:
This PR adds tooltips for CU lookup pages.

This required editing our tooltips JSON file so that it also takes the base OpenAPI template into account. As CU lookup is not a resource, it also required some custom handling in the frontend.

### DIFF
--- a/frontend/src/tooltip/tooltips.json
+++ b/frontend/src/tooltip/tooltips.json
@@ -153,5 +153,25 @@
         "type": "The type of the notice.",
         "source": "The URI of the resource concerned by the event.",
         "data": "The data of the notice."
-    }
+    },
+    "error_message": {
+        "code": "The error code.",
+        "details": "Detailed information about the error.",
+        "hint": "A hint to help resolve the error.",
+        "message": "The error message."
+    },
+    "controllable_unit_lookup_request": {
+        "end_user": "Birth number or organisation number of the end user.",
+        "controllable_unit": "The business ID of the controllable unit to lookup.",
+        "accounting_point": "The accounting point ID of the controllable unit(s) to lookup. `GSRN` metering point id."
+    },
+    "controllable_unit_lookup_response_item": {
+        "id": "The surrogate key of the controllable unit.",
+        "business_id": "The business ID of the controllable unit.",
+        "name": "The name of the controllable unit.",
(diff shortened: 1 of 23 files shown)


frontend/src/tooltip/tooltips.json
```diff
@@ -153,5 +153,25 @@
...
+    },
+    "controllable_unit_lookup_response_item": {
+        "id": "The surrogate key of the controllable unit.",
+        "business_id": "The business ID of the controllable unit.",
+        "name": "The name of the controllable unit.",
+        "accounting_point_id": "The accounting point ID of the controllable unit.",
+        "end_user_id": "The ID of the current end user on the controllable unit.",
+        "technical_resources": "The technical resources belonging to the controllable unit."
+    },
+    "controllable_unit_lookup_response": null
```
enzo-crance-statnett, line 176: We could add an extra filter to remove the null value here. We don't use this entry so it is not a major issue, but in a 100% clean solution it should not be there.

openapi/openapi-api-base.yml
enzo-crance-statnett, line 1: Here the diff is big but no details change on the fields, we are just moving the content of the array schema into its separate schema so we can access it while generating `tooltips.json`.

enzo-crance-statnett (commented):
The work on the custom stack components made me think that ideally they should be split into 2 :

- stack components adding auth/permissions stuff
- stack components adding tooltips

The auth ones would rely on the tooltips ones. I tried doing that but the `Children.map` messes up the props, so we cannot run it twice in the current state. It needs a bit more thinking and I evaluated it a bit too much for the current ticket.

frontend/src/tooltip/tooltips.json
krilor: Shouldn't we remove the "_item" bit? Like, we don't really have "_item" on the other things that return lists. So why do we want that here?

E.g. controllable unit: https://github.com/elhub/flex-information-system/blob/6c0e76be71db3bbed3406f177c95fedd1a365d18/openapi/openapi-api.json#L2622-L2635

krilor (approved):
Consider the rename to remove "_item" , but otherwise looks good.

PR #93, opened 2025-03-17
feat: remove end user ID from CU lookup response (FLEX-511)

enzo-crance-statnett:
This PR removes the `end_user_id` field of the CU lookup response.

First, it is unnecessary, because the user doing the lookup already knows their business ID, so they are likely to know their name. Second, it is a technical ID that does not mean much functionally, and if the user launching the lookup is not allowed to see the associated party, then they cannot even get the name from this ID.

Rød PR altså :wink:


krilor (approved):
Red PRs are the best PRs 🥳

PR #96, opened 2025-03-18
Disable lint in npmVerify

johannes-gs:
## 📋 Checklist

* ✅ Lint checks passed on local machine.
* ⚠️ **No tests could be run for this PR.**


johannes-gs:
The linting step in npmVerify is superfluous, as we now lint during PR creation with megalinter. It will likely be remove in the latest build-configuration version, and then this flag can be removed as well.

PR #98, opened 2025-03-18
feat: even more agressive expiry in portal (FLEX-512)

krilor:


krilor:
Setting this makes us rely soley on Etag, which is what we want currently. _If_ we ever see that this becomes a problem/load, then we can set different expiry on different file types.

FYI: I've been using Firefox and it seems firefox and Chrome have different behaviour here. Ref
https://stackoverflow.com/questions/43659756/chrome-ignores-the-etag-header-and-just-uses-the-in-memory-cache-disk-cache

enzo-crance-statnett (approved):

PR #103, opened 2025-03-18
feat: client subresource for entity (FLEX-520)

enzo-crance-statnett:
This PR adds a client resource to the database and shows it as a subresource for entities in the UI.

List on entity show page

Screenshot 2025-03-18 154559

Client show

Screenshot 2025-03-18 154924

Client create/edit

Screenshot 2025-03-18 154938


local/scripts/openapi_to_db.py
enzo-crance-statnett, line 1: History tables were generated when we just need audit (bug). Actually it also shows another bug : if we don't have history, we don't use `flex.versioning` so `recorded_at` never changes. We need to think a bit more about resources that have audit _but no history_, because we don't have many of them (there is notification as well for instance).

krilor: Yeah, lets fix this in a separate story for notification. Can you add a story? There is also a bit of a mixup with words here as generally history is what enables audit in my head. Just tracking the latest change isn't really providing audit as such IMHO. We lack documentation about this as well.

For this specific resource we should enable history imho but not expose it in the API. Knowing when and who changed credentials is a nice thing to be able to provide history on IMHO.

krilor (changes requested):
This is looking very good! Not saying that all the comments I made must result in change, but we should at least consider them. The _history one is a must tho.

db/flex/entity_client.sql
```diff
@@ -0,0 +1,48 @@
+CREATE TABLE IF NOT EXISTS entity_client (
+    id bigint PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+    entity_id bigint NOT NULL REFERENCES entity (id),
+    client_id uuid NOT NULL UNIQUE DEFAULT (public.uuid_generate_v4()),
+    name text NULL CHECK (char_length(name) <= 64),
```
krilor, line 5: Should we make this a bit longer just in case? E.g. 4x

openapi/resources.yml
```diff
@@ -531,6 +531,75 @@ resources:
...
+        description: Reference to the entity that this client is attached to.
+        format: bigint
+        type: integer
+        x-no-update: true
+        x-foreign-key:
+          resource: entity
+          field: id
+        example: 30
+      name:
+        description: Name of the client. Maximum 64 characters.
```
krilor, line 565: NIT: It feels a bit "off" to put the maxLength in the description like this.

krilor (approved):

PR #105, opened 2025-03-19
feat: pg_timeline in separate schema (FLEX-481) (FLEX-493)

krilor:


db/flex_base.sql
```diff
@@ -5,8 +5,14 @@
 SET search_path TO flex, public;
 
 \i temporal_tables/temporal_tables.sql
+
 \i pg_timeline/pg_timeline.sql
+GRANT USAGE ON SCHEMA timeline TO flex_anonymous;
```
krilor, line 10: This part of the db bits is just temporary until we run liquibase as part of actual deploy

db/gs1/gs1.sql
```diff
@@ -1,9 +1,9 @@
 --liquibase formatted sql
 
--- changeset gs1-create-schema:1
+-- changeset flex:gs1-create-schema
```
krilor, line 3: so the format is `<author>:<id>` so this makes more sense ;)

PR #109, opened 2025-03-19
feat(auth): use entity client resource in login (FLEX-521)

enzo-crance-statnett:
This PR makes use of the entity client resource in the actual authentication routes that were previously using client information stored in entity. Client credentials and JWT Bearer authentication methods are updated.


enzo-crance-statnett:
Noticed both the issuer and subject claims of the token were represented using the same custom type. As they contain different values, the parsing function becomes a bit complex. It is now split into 2 types to make things simpler and have better type safety.

krilor (changes requested):
As we discussed: Remove prefix on iss

krilor (approved):

PR #110, opened 2025-03-21
docs: status transitions (FLEX-532)

kelvestad:
Moved status transitions from processes to resources. Corrected small spelling error in Revocation process. Changed status Pending to Requested in "Status transitions for grid prequalification" according to status in SPG grid validation resource.


enzo-crance-statnett (approved):

JanMStrand (commented):
This looks really good.
NCDR will in phase 2 do the status transitions for different processes. Status by now (hehe) is that they will use codes for statuses and of course a table explaining....

PR #111, opened 2025-03-21
docs: clean up ids (FLEX-552)

kelvestad:


enzo-crance-statnett (dismissed):

docs/diagrams/service_provider_switching.plantuml
```diff
@@ -23,8 +23,7 @@ activate SP_NEW
 SP_NEW -> FIS: Register CU service provider contract
 note left
     Data:
-    - service provider id
-    - controllable unit id
+    - controllable unit
```
krilor, line 26: NIT: when calling the API, the SP must also add their own ID as part of the data. This is pre-filled in the UI

krilor (approved):

docs/diagrams/service_provider_switching.plantuml
kelvestad: ah, then Ill add it back!

krilor (approved):

PR #122, opened 2025-03-28
refactor: remove client information on entity resource (FLEX-522)

enzo-crance-statnett:
Now that entity clients have been added and auth functions got updated to use them, client information on the entity resource becomes useless.

This PR therefore mostly deletes code.


openapi/resources.yml
```diff
@@ -459,8 +459,7 @@ resources:
     operations:
       - list
       - read
-      - update
-    generate_views: false
+    generate_views: true
```
enzo-crance-statnett, line 462: As there is nothing updatable left on the resource, better remove this operation and make the view automated again :tada:

enzo-crance-statnett:
I kept the e-mail format as an input to `add_test_account` for compatibility with the other repository.

enzo-crance-statnett:
It looks like person entities are always created with a `pid` business ID, organisation entities are always created with an `org` business ID, and users have no create/update authorisation on the entity resource. This means there is now a direct link between `entity.business_id_type` and `entity.type`. Should we tie these fields together, with a constraint or by generating one of them from the other ?

krilor (approved):

krilor:
> I kept the e-mail format as an input to `add_test_account` for compatibility with the other repository.

krilor:
> It looks like person entities are always created with a `pid` business ID, organisation entities are always created with an `org` business ID, and users have no create/update authorisation on the entity resource. This means there is now a direct link between `entity.business_id_type` and `entity.type`. Should we tie these fields together, with a constraint or by generating one of them from the other ?

Good ideas. Lets consider these things when we do Entity CREATE aka https://elhub.atlassian.net/browse/FLEX-176

PR #123, opened 2025-03-28
feat(db): allow multiple SP product application per SP & SO (FLEX-555)

enzo-crance-statnett:
This PR allows an SP to create several product applications for the same SO. There is actually nothing else to do than lifting a constraint.

No additional constraint on the actual product types was added for now, but we could error if a product type was already applied for in a previous application.


krilor (approved):
My favorite type of "feat"! Tiny change and actually just deleting stuff 🤩

PR #125, opened 2025-03-28
feat(db): restrict products update on SPPA as SP to requested status (FLEX-556)

enzo-crance-statnett:
The main addition of this PR is a trigger restricting the product type list update for SP, so that it becomes impossible after the status is changed from `requested`.

However, this change implies some modifications/simplifications :

- As the SP can only edit the list before the status changes, they must be careful so that the application does not jump automatically to `communication_test` if they input a list where they have been qualified for all the product types. Therefore, this PR _removes_ the automated status update.

  Note that this does not really impact the SO. They just won't know as directly whether all product types have been qualified before, but they can already know that with the API thanks to SPPA-SO001. We have an update for that planned as another story ([FLEX-562](https://elhub.atlassian.net/browse/FLEX-562)), but it is more of a quality of life improvement.

- As the SP cannot edit the product types after the status moves, the other automated status update in the status transition diagram is outdated. This means that `requested` is now more of an initial state only. We now enforce it in the database and the process allows back-and-forth transitions between `not_qualified` and `in_progress` instead.

[FLEX-562]: https://elhub.atlassian.net/browse/FLEX-562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


db/flex/service_provider_product_application.sql
```diff
@@ -106,6 +106,38 @@ FOR EACH ROW
 EXECUTE FUNCTION
 service_provider_product_application_product_type_ids_insert();
 
+-- trigger to reject updates by SP to the product types if the status of the
+-- application is no longer `requested`
+
+CREATE OR REPLACE FUNCTION
+service_provider_product_application_product_type_ids_update_only_requested()
```
krilor, line 113: NIT: a tiny future proofing for change could be to omit the "_only_requested" part.

krilor (changes requested):
Looks very good. But we must add docs.

db/flex/service_provider_product_application.sql
enzo-crance-statnett: `service_provider_product_application_product_type_ids_update` already exists. Is it better to merge both functions ?

enzo-crance-statnett:
The pre-existing logic allows jumping automatically to `communication_test` if the SP has already been qualified once for each product type they are applying for. This means that the "update while requested" feature is not available anymore if they update their product list to a list full of already qualified product types.

enzo-crance-statnett:
Updated the docs and implemented the simplifications based on our discussions. Also updated the PR description accordingly.

enzo-crance-statnett:
https://elhub.atlassian.net/browse/FLEX-563 also identified as another quality of life improvement.

krilor (approved):

PR #129, opened 2025-04-01
docs: improve resource docs (mainly validation rules and status) (FLEX-557)

enzo-crance-statnett:
This PR improves documentation on our resources. It removes the sub-categories on validation rules and makes sure all rules related to status fields are explained. It also makes unicity/multiplicity explicit on resources representing relations.

Opening for information, but as a draft, because there is still something to do with SPG grid prequalification.


enzo-crance-statnett:
There was a divergence between the SPG-GP status diagram and the rules in the database. After discussion, we keep `requested` as an initial state only for now, and we update the diagram (in other words, we fix the diagram but the database is right, not the other way around).

docs/resources/service_providing_group_product_application.md
```diff
@@ -35,7 +45,10 @@ type, for the SPG to deliver a product to the SO later.
 
 ## Validation Rules
 
-No validation rules.
+| Validation rule key | Validation rule                                                   | Status |
+|---------------------|-------------------------------------------------------------------|--------|
+| SPGPA-VAL001        | Product applications can only be created on active SPGs.          | DONE   |
+| SPGPA-VAL002        | Product applications can only be created on active product types. | DONE   |
```
krilor, line 51: Should we do:

```suggestion
| SPGPA-VAL002        | Product applications can only be created on active system operator product types. | DONE   |
```
?

krilor (approved):
🤩 Huge improvement. I love it.

PR #130, opened 2025-04-02
refactor(db): make history_audit ready for liquibase

krilor:


requirements-dev.txt
```diff
@@ -1,15 +1,16 @@
+argon2-cffi
+click
+cryptography
 hypothesis
+Jinja2
```
krilor, line 5: just adding Jinja2. The rest is just sorting.

krilor:
This PR changed from "move stuff to new `logic` schema" to "add liquibase stuff to the history_audit" templating since i learned that neither policies nor triggers can actually live in another schema ..

I'm currently rethinking the strategy of the "logic" schema, but this change should be safe to merge as is and is a step in the direction of liquibase-for-flex

enzo-crance-statnett (approved):

PR #131, opened 2025-04-02
feat(backend): restrict login with client credentials method (FLEX-480)

enzo-crance-statnett:
This PR adds restrictions to _failed_ login attempts with the client credentials method.

I gathered ideas from our discussions and also from other sources, primarily the [OWASP recommendations](https://owasp.org/www-community/controls/Blocking_Brute_Force_Attacks).

The 2 main features are _delay on failed login_ and _maximum number of failed login attempts in a time window_.

- **Delay on failed login**
  We should not lock accounts, because among other issues it makes us vulnerable to denial of service attacks. Delaying does not prevent bots from trying to log in, but it makes brute force attacks much slower while not really annoying the real user (waiting for a few extra seconds does no harm, especially if only on failed attempts). I implemented it with an environment variable so that we can disable it when testing.

- **Max failed login attempts**
  Regarding spamming from a same IP, it can be a problem if we ban an IP of an ISP for example. A better solution would be to add the client IP in entity clients when they add their passwords, so we can switch from blacklisting to whitelisting. But this is more complex, and I went for the simpler way at first. We allow 20 failed login attempts per IP per hour. No permanent ban but they then have to wait until the end of the window to retry.


backend/auth/auth.go
```diff
@@ -32,28 +32,69 @@ const (
...
+func readUserIP(r *http.Request) ipAddress {
+	ip := r.Header.Get("X-Real-IP")
+	if ip == "" {
+		ip = r.Header.Get("X-Forwarded-For")
+	}
+	if ip == "" {
+		ip = r.RemoteAddr
+	}
+	return ipAddress(ip)
+}
```
enzo-crance-statnett, line 60: > [!NOTE]
> As we trust the reverse proxy (nginx), we should probably just check `X-Forwarded-For` there...

backend/flex.go
```diff
@@ -188,6 +189,17 @@ func Run(ctx context.Context, lookupenv func(string) (string, bool)) error { //n
 		return fmt.Errorf("%w: FLEX_OIDC_POST_LOGOUT_REDIRECT_URL", errMissingEnv)
 	}
 
+	loginDelay := 2 * time.Second
```
enzo-crance-statnett, line 192: Used a 2-second delay by default.

test/auth_tests/test_client_credentials.py
enzo-crance-statnett, line 1: > [!NOTE]
> We could parameterise the failed login time window as well, so that we can set it to a few seconds in the tests, get the `HTTP 429`, wait for the window to pass, then retry and check we get a `HTTP 400`.

backend/auth/auth.go
```diff
@@ -1014,6 +1055,37 @@ func (auth *API) clientCredentialsHandler( //nolint:funlen
...
+				ctx, "client credentials failed login window reset", "client", clientIP,
+			)
+			record = failedLoginRecord{
+				attempts:     0,
+				firstAttempt: time.Now(),
+			}
+		}
+
+		if record.attempts >= auth.maxFailedAttemptsInLoginWindow {
+			ctx.AbortWithStatusJSON(http.StatusTooManyRequests, oauthErrorMessage{
```
enzo-crance-statnett, line 1077: The status here requires an update to the auth API docs...

test/auth_tests/test_client_credentials.py
enzo-crance-statnett: Done !

backend/auth/auth.go
enzo-crance-statnett: Done !

krilor (commented):
Lets talk about this today. This is generally looking good, but there are some things I would like to talk about.

1. Real user IP: Doesn't it make sense to do this as an early middleware similar to the [chi one](https://github.com/go-chi/chi/blob/master/middleware/realip.go). This is helpful also for other things, like security logging. We also need to consider the use of headers very carefully and ensure that our load balancer and reverse proxy act as we assume wrt setting headers. This topic made me remember [this CVE](https://wpscan.com/vulnerability/1d748f91-773b-49d6-8f68-a27d397713c3/) which is highly relevant for us to mitigate ;)
2. Configuration: Do we really need to make this configurable? "Security misconfiguration" is on the OWASP Top 10
3. The current implementation is using custom rate limiting logic. Should we rather use https://pkg.go.dev/golang.org/x/time/rate ?
4. Concurrent access to map. Concurrent access to map have AFAIK undefined behaviour. We should look at adding some mutexes or sync.Map
5. Testability. If we factor this addition a bit differently, would it be easier to test? Are there any parts of it that should have a _test.go?

backend/auth/auth.go
```diff
@@ -1014,6 +1056,37 @@ func (auth *API) clientCredentialsHandler( //nolint:funlen
 		ctx, tx, payload.ClientID, payload.ClientSecret,
 	)
 	if err != nil {
+		// it is a failed login attempt
```
krilor, line 1059: Is something a bit off here? Shouldn't we check if rate limit was exceeded _before_ trying/doing the login?

My head might be a bit slow today, but right now it seems like we are just using the rate limit to return different HTTP statuses (?)

enzo-crance-statnett: I thought we said that we should rate-limit _failed_ queries only, and we cannot know if it's a failed query until this moment in the code. But maybe I understood it in a wrong way, and as we already have the delay, we could very well rate-limit the whole thing. 🤔

enzo-crance-statnett:
Definitely agree on points 1, 3 and 4. For points 2 and 5 I'd like to talk to make sure I understand your approach.

enzo-crance-statnett:
Updated the solution to cover most of what we lacked in the first draft.

The IP address retrieval is moved to a stdlib middleware expecting `X-Real-IP` to be defined, which is the case locally with nginx, and we need to check it for the test environment.

The solution of the stdlib `time/rate` package is based on [token buckets](https://en.wikipedia.org/wiki/Token_bucket), so our previous way to measure the limit ("20 logins per hour") does not really apply anymore. Instead, we now have 20 tokens available at the beginning, and a new one is generated every 3 minutes. This means a user can try 20 logins in one shot, but then the bucket will be empty and they will have to wait for 3 minutes. If they wait 3 minutes every time they fail a login, it means in total they can run 40 logins in an hour. But I think it remains acceptable, the goal there is just to have a limit that remains sensible.

enzo-crance-statnett:
The technical thing there is that we wanted to make sure the rate limiter applies to _failed_ logins only.

The rate limiter library offers three main methods : `Allow`, `Reserve` and `Wait`. `Wait` blocks until a token is available, and we want no blocking behaviour. `Allow` was the solution we would have taken if valid logins were to be limited as well, but here on a valid login we want to give the token back to the rate limiter. However, `Allow` internally takes a token but just returns a boolean and no way to give it back. We are left with `Reserve` that takes the token and returns a reservation object that we can use for various things, in particular it has this `Cancel` method that gives the token back to the rate limiter.

The problem is that currently, a high number of _successful_ logins depletes the token pool, even though it is just temporary (after login the token is given back), so the test fails. At this point, I am wondering if we really need to care about not restricting successful logins. Using `Allow` would be simpler, and anyway successful logins are supposed to be less spammy ?

krilor (commented):
> The problem is that currently, a high number of successful logins depletes the token pool, even though it is just temporary (after login the token is given back), so the test fails.

This seems weird. Are we running tests in parallell? I was assuming that our tests are sequential and given a correct implementation the token should be released even before the next auth request? Lets look at the the behaviour together tomorrow.

enzo-crance-statnett:
As it is not straightforward to properly restore tokens with `time/rate`, rolling back to a manual solution with exponential delays. Tested.

krilor (changes requested):
We need to think a bit more about the reset logic, as we discussed yesterday

krilor:
We are pausing this a bit while finishing some other stuff. Marked as draft to make it evident that this is not proposed for merge in its current state.

PR #133, opened 2025-04-02
feat(db): add SPG-GP for the ISO of the CU on SPG membership insert (FLEX-564)

enzo-crance-statnett:
This PR adds a SPG grid prequalification resource for each CU added to a SPG.

Pausing work on this as it raised some questions. Draft until we clarify (also it breaks tests for now).

### DIFF
--- a/db/flex/service_providing_group.sql
+++ b/db/flex/service_providing_group.sql
@@ -36,34 +36,52 @@ FOR EACH ROW
 WHEN (OLD.status IS DISTINCT FROM NEW.status) -- noqa
 EXECUTE FUNCTION status_update('new');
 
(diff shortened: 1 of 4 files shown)


krilor:
> Pausing work on this as it raised some questions.

Which questions?

Also: Should we consider putting the "ensure all relevant ISOs have SPG GP" functionality in a common function that is called by triggers on both tables?

enzo-crance-statnett:
Put the functionality into a common function. Need to fix tests before it's ready.

enzo-crance-statnett:
The following behaviour is implemented and tests are green :

SPG grid prequalification is created
- on SPGM insert, if the membership is valid when inserted ;
- on SPG activation, for memberships valid at the time of activation.

krilor (approved):

PR #137, opened 2025-04-08
feat(ui): allow filtering of entities and parties (FLEX-568)

enzo-crance-statnett:
Filtering on entity, type and status for party : 

Screenshot 2025-04-08 155425

Filtering on type for entity :

Screenshot 2025-04-08 155713

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -7242,9 +7242,9 @@
             "license": "(WTFPL OR MIT)"
         },
         "node_modules/vite": {
-            "version": "6.2.3",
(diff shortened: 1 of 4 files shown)


enzo-crance-statnett:
cc @kelvestad

krilor (approved):
Nice!

But aren't we missing the most important filter - on name?

Also a quickfilter to show just the current entity/party makes sense to me

enzo-crance-statnett:
Name filter OK, but the `QuickFilter` interacts badly with it. Anyway, I used an `AutocompleteReferenceInput`, so only the first letters of the name are needed to find it.

Not a `QuickFilter`, but still a quick filter...

PR #143, opened 2025-04-15
feat(backend): notify SP on implicit CUSP termination (FLEX-531)

enzo-crance-statnett:
This PR adds a notification for SP on implicit CUSP termination caused by end user termination.

I narrowed down the scope to the specific case of _termination_ : an _infinite_ CUSP (open-ended valid time), initially allowed by an infinite end user contract, gets implicitly terminated because an end was set to the valid time of the end user contract.

Indeed, there is a range of other cases of inconsistency that can be brought by an update to the end user timeline (basically any situation where the new timeline does not cover the whole CUSP contract), but these are not necessarily _terminations_ (_e.g._, start valid time moves forward in time so the beginning of the CUSP is no longer covered -- it is not a termination), and they could be associated to the lighter _notice_ instead of _notification_ ?

Not automatically tested, because accounting point end user is not available on the API. We do not yet have an easy way to test system updates to internal data. Tested manually by editing AP-EU and checking the right notification got generated.

### DIFF
--- a/backend/event/worker.go
+++ b/backend/event/worker.go
@@ -87,7 +87,7 @@ func (eventWorker *Worker) Stop(ctx context.Context) error {
 // from a replication connection dedicated to event processing. It has access
 // to a transaction to perform operations on the database.
 //
-//nolint:cyclop
+//nolint:cyclop,funlen
 func (eventWorker *Worker) handleMessage(
 	ctx context.Context,
 	message *pgrepl.Message,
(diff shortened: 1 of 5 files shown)


backend/event/worker.go
```diff
@@ -119,9 +119,14 @@ func (eventWorker *Worker) handleMessage(
...
-		if err != nil {
-			return fmt.Errorf("could not get party of identity: %w", err)
+		var eventPartyID int
+		if event.RecordedBy == 0 { // system event
+			eventPartyID = 0
+		} else {
+			eventPartyID, err = authModels.PartyOfIdentity(ctx, tx, event.RecordedBy)
+			if err != nil {
+				return fmt.Errorf("could not get party of identity: %w", err)
+			}
```
enzo-crance-statnett, line 129: This allows the event worker not to crash when it encounters an event generated by the system, which is the case here. Indeed, as we can't track who updated the end user timeline externally, we consider the implicit termination caused by the system (`event.recorded_by = 0`).

krilor:
I've converted this to a draft since we want to discuss how we make end user, energy supplier and brp available in the api first. If EU is available directly on the api, then the event in question here should be related to that resource.

enzo-crance-statnett:
The current decision is to provide this information as a _general inconsistency notice_ rather than a _specific notification for termination_. Addition of notifications will be re-evaluated later.

Closed in favour of https://github.com/elhub/flex-information-system/pull/197.

PR #144, opened 2025-04-15
feat: open AP-BRP and AP-ES resources in the data API (FLEX-490)

enzo-crance-statnett:
This PR makes BRP and ES information on accounting points available in the data API.

Draft because tests are missing. Also some information may be needed in the UI, on CU show pages.

Not sure about the history part, these resources have history enabled but as the data is immutable and will come from Elhub in the future, maybe we don't care about keeping history in the API. Could be just an internal thing.

An alternative to opening these resources was to edit the CU view to add 2 extra fields showing the last BRP and ES to the current date on the AP. That was however making the view a bit complex, and if several BRPs or ESs are there during a single CU record time, just showing the latest one makes the CU history table a bit unreliable.

Also, as we had these resources started internally, it made sense to just finish and open them instead of adding the data in several places. Especially in the context where the AP resource was already partially opened and more and more things are gravitating around it.

RLS policies to discuss afterwards.


enzo-crance-statnett:
After discussion, covered in a different way (non-materialised resource) in https://github.com/elhub/flex-information-system/pull/147.

Most of the changes here were used as input to the new PR. Closing the former one.

PR #145, opened 2025-04-25
feat(db): add metering grid area resource (FLEX-487)

enzo-crance-statnett:
Combination of https://github.com/elhub/flex-information-system/pull/139 and https://github.com/elhub/flex-information-system/pull/140, this time complete and targeted towards `main`.


db/flex/accounting_point.sql
```diff
@@ -1,12 +1,10 @@
...
     id bigint PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
-    business_id text UNIQUE NOT NULL,
-    balance_responsible_id bigint NULL,
-    balance_responsible_party_type text GENERATED ALWAYS AS (
-        'balance_responsible_party'
-    ) STORED,
+    business_id text UNIQUE NOT NULL CHECK (
+        validate_business_id(business_id, 'gsrn')
+    ),
+    metering_grid_area_id bigint NOT NULL,
```
krilor, line 7: ap mga is time dependent, so we cannot have it as a field like this.

krilor (changes requested):
Whats the reasoning for exposing the `metering_grid_area` as part of this PR?
As discussed earlier this week, multiple fields/relations on the MGA is actually time-dependent (SO and price area). We said it was ok with the suggested, non time-dependent, structure to begin with _since it was internal_.

Now that is is suggested to be external we actually need to consider it much more carefully. Do we really need to make it external? What is the current use case for it?

enzo-crance-statnett:
My reasoning was that technical IDs we show in the API should be explainable, otherwise we should return other information instead (e.g. business ID). This is the case with MGA ID on the AP resource here. Without MGA open, it's just a meaningless technical ID. So I went with an open MGA resource as well, without too much thinking since it is public data.

If we turn AP-MGA into a time-dependent relation (so as another resource), then the AP resource does no longer contain any such "meaningless" field and we are fine with keeping it closed for now. So fixing one issue also removes the other.

db/flex/accounting_point_rls.sql
```diff
@@ -1,9 +1,9 @@
--- We shouldn't really need RLS, but keeping the filename consistent with the other tables
-GRANT SELECT ON flex.accounting_point TO flex_common;
+ALTER TABLE IF EXISTS accounting_point ENABLE ROW LEVEL SECURITY;
 
--- internal
-GRANT SELECT ON accounting_point TO flex_internal_event_notification;
-CREATE POLICY "AP_INTERNAL_EVENT_NOTIFICATION" ON accounting_point
```
krilor, line 6: Need to double check that we can remove this

enzo-crance-statnett: Looks like we cannot, indeed. The AP table is used in the event worker to identify notification recipients.

krilor (approved):

PR #146, opened 2025-04-28
feat(db): add retailer balance responsibility as internal resource (FLEX-486)

enzo-crance-statnett:
Links MGA + ES + direction to a BRP for a given valid time.


db/flex/retailer_balance_responsibility.sql
```diff
@@ -0,0 +1,59 @@
+-- internal
+-- This table stores the balance responsible party chosen by each energy
+-- supplier in the different metering grid areas, in both possible directions.
+CREATE TABLE IF NOT EXISTS retailer_balance_responsibility (
```
krilor, line 4: Shouldn't we make this energy supplier balance responsibility to fit with "our" way of naming things? The external (eSett) name is not part of our "bounded context".

db/flex/retailer_balance_responsibility.sql
```diff
@@ -0,0 +1,59 @@
...
+    energy_supplier_party_type text GENERATED ALWAYS AS (
+        'energy_supplier'
+    ) STORED,
+    balance_responsible_party_id bigint NOT NULL,
+    balance_responsible_party_party_type text GENERATED ALWAYS AS (
+        'balance_responsible_party'
+    ) STORED,
+    direction text NOT NULL CHECK (
+        direction IN (
+            'up',
```
krilor, line 17: direction in this context should be understood as "energy direction" as we see in https://opendata.esett.com/rbr

So we should probably rename to "energy_direction" and use "consumption", "production" for clarity ?

krilor (changes requested):
A couple tiny adjustments and we are good to go :)

PR #147, opened 2025-04-30
feat: accounting point balance responsible party (FLEX-490)

enzo-crance-statnett:
This PR adds an AP-BRP _pseudo_-resource (with non-stable IDs and no backing table). Information comes from the recently added AP-MGA and ES-BRP read-only resources, joined with AP-ES.


db/api/accounting_point_balance_responsible_party.sql
```diff
@@ -2,13 +2,50 @@
...
         accounting_point_id,
         balance_responsible_party_id,
-        valid_time_range,
-        record_time_range,
-        recorded_by
-    FROM flex.accounting_point_balance_responsible_party
+        direction,
+        row_number() OVER () AS id,
+        lower(valid_time_inter) AS valid_from,
+        upper(valid_time_inter) AS valid_to
```
krilor, line 10: We discussed the idea of limiting the view here if the cusp range is smaller than the ap_brp range. Must not forget to do that.

db/api/accounting_point_balance_responsible_party.sql
```diff
@@ -2,13 +2,50 @@
...
+    FROM ( -- noqa
+        SELECT
+            ap_es.accounting_point_id,
+            es_br.balance_responsible_party_id,
+            es_br.direction,
+            (
+                ap_es.valid_time_range
+                * ap_mga.valid_time_range
+                * es_br.valid_time_range
+            ) AS valid_time_inter
```
krilor, line 20: NIT: consider renaming `valid_time_inter` -> `valid_time_range` for consistency?

db/api/accounting_point_balance_responsible_party.sql
```diff
@@ -2,13 +2,50 @@
...
 WITH (security_invoker = true) AS (
     SELECT
-        id,
         accounting_point_id,
         balance_responsible_party_id,
-        valid_time_range,
-        record_time_range,
-        recorded_by
-    FROM flex.accounting_point_balance_responsible_party
+        direction,
```
krilor, line 7: Ref https://github.com/elhub/flex-information-system/pull/146#discussion_r2067931375

We really should denote this as "energy_direction" to make it clear that this is something different from "regulation direction". The term energy direction is from https://opendata.esett.com/rbr BTW

docs/resources/accounting_point_balance_responsible_party.md
```diff
@@ -0,0 +1,99 @@
...
+|----------------|-----------------|--------|
+| APBRP-FISO001  | Read all APBRP. | DONE   |
+
+#### Market Operator
+
+No policies.
+
+#### System Operator
+
+No policies.
```
krilor, line 76: SO should see "their own"?

backend/data/data.go
```diff
@@ -56,6 +56,8 @@ func NewAPIHandler(
 	// all other requests are forwarded to PostgREST
 	mux.HandleFunc("/accounting_point", data.postgRESTHandler)
 	mux.HandleFunc("/accounting_point/{path...}", data.postgRESTHandler)
+	mux.HandleFunc("/accounting_point_balance_responsible_party", data.postgRESTHandler)
+	mux.HandleFunc("/accounting_point_balance_responsible_party/{path...}", data.postgRESTHandler)
```
krilor, line 60: Since IDs are unstable and not to be used. The singular endpoint is not something we want/need. This resource should just support "list"

db/api/accounting_point_balance_responsible_party.sql
```diff
@@ -2,13 +2,50 @@
...
+        FROM flex.accounting_point_energy_supplier AS ap_es -- noqa
+            INNER JOIN flex.accounting_point_metering_grid_area AS ap_mga
+                ON ap_mga.accounting_point_id = ap_es.accounting_point_id
+            INNER JOIN flex.energy_supplier_balance_responsibility AS es_br
+                ON es_br.metering_grid_area_id = ap_mga.metering_grid_area_id
+        WHERE
+            es_br.energy_supplier_id = ap_es.energy_supplier_id
+            AND ap_es.valid_time_range && ap_mga.valid_time_range
+            AND ap_es.valid_time_range && es_br.valid_time_range
+    ) AS ap_brp
```
krilor, line 30: Should we consider replacing https://github.com/elhub/flex-information-system/blob/main/db/flex/accounting_point_balance_responsible_party.sql with this (as a view) ?

In any case we must clean out the old stuff as part of the work we are doing now.

enzo-crance-statnett: Indeed, we had a trivial approximation of AP-BRP before in the `flex` schema. I removed it but it broke a test for technical resource. I'll do another pass on the tests once I manage to implement the constraints for SP.

enzo-crance-statnett:
Still draft because as we replaced the former BRP resource, it breaks a few tests on other resources. Also adding CUs in the other tests breaks the logic of the current AP-BRP test...

db/authz/accounting_point_service_provider.sql
```diff
@@ -0,0 +1,18 @@
+CREATE OR REPLACE VIEW accounting_point_service_provider AS (
+    SELECT
+        ap.id AS accounting_point_id,
+        cusp.service_provider_id,
+        range_agg(cusp.valid_time_range) AS valid_timeline
+    FROM flex.controllable_unit_service_provider AS cusp -- noqa
+        INNER JOIN flex.controllable_unit AS cu
+            ON cusp.controllable_unit_id = cu.id
+        INNER JOIN flex.accounting_point AS ap
+            ON ap.business_id = cu.accounting_point_id
```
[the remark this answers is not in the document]
enzo-crance-statnett: These records won't have any effect, but it does not hurt to filter them out. Fixed.

db/flex/accounting_point_balance_responsible_party.sql
```diff
@@ -1,38 +1,28 @@
...
+            ap_es.valid_time_range
+            * ap_mga.valid_time_range
+            * es_br.valid_time_range
+        ) AS valid_time_range
+    FROM flex.accounting_point_energy_supplier AS ap_es -- noqa
+        INNER JOIN flex.accounting_point_metering_grid_area AS ap_mga
+            ON ap_mga.accounting_point_id = ap_es.accounting_point_id
+        INNER JOIN flex.energy_supplier_balance_responsibility AS es_br
+            ON es_br.metering_grid_area_id = ap_mga.metering_grid_area_id
+    WHERE
```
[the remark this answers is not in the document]
enzo-crance-statnett: TIL you can put an arbitrary boolean there and not only one equality between two fields 😄 I suppose it makes the view even more efficient.

db/authz/accounting_point_service_provider.sql
```diff
@@ -0,0 +1,18 @@
+CREATE OR REPLACE VIEW accounting_point_service_provider AS (
+    SELECT
+        ap.id AS accounting_point_id,
+        cusp.service_provider_id,
+        range_agg(cusp.valid_time_range) AS valid_timeline
```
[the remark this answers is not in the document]
enzo-crance-statnett: This is inheritance from the former implementation, indeed no reason to keep that. We can switch back to aggregating where the view is used, not in its definition. 👍🏻

enzo-crance-statnett:
TR test not fully fixed. I suspect an overlap in the manually inserted record time ranges in the test data. It's the last thing to fix here hopefully.

db/api/accounting_point_balance_responsible_party.sql
```diff
@@ -4,19 +4,65 @@
...
+    WHERE current_role = 'flex_system_operator'
+        AND ap.system_operator_id = flex.current_party()
+    -- RLS: APBRP-SP001
+    UNION ALL
+    SELECT
+        id,
+        accounting_point_id,
+        balance_responsible_party_id,
+        energy_direction,
+        CASE WHEN lower(valid_time_range) < current_timestamp + '2w'::interval
```
krilor, line 38: When the record is past the window, remove the row.

krilor (approved):

PR #148, opened 2025-05-05
chore: api schema in liquibase (FLEX-502)

krilor:


justfile
```diff
@@ -28,7 +28,7 @@ unload:
 
 
 # load the database
-load: liquibase
+load:
```
krilor, line 31: This change is intentional btw. We actually do not need liquibase to run on local. And with the suggested change it breaks since flex is still not liquibase-managed (and api depends on flex).

So just removed this to make this change to api independent from the rest of the changes.

enzo-crance-statnett (approved):

PR #162, opened 2025-05-08
feat(backend): add delayer to client credentials login (FLEX-480)

enzo-crance-statnett:
Refreshed version of https://github.com/elhub/flex-information-system/pull/131 after the latest discussion.


enzo-crance-statnett:
Last thing to (possibly) add is possibly a feature toggle, we can implement it for example by

- adding a new environment variable as a flag OR
- turning the existing one into an enum (`test`/`off` for instance) OR
- changing the default values to something very permissive until we complete the real IP task.

backend/auth/iplogindelayer.go
```diff
@@ -0,0 +1,166 @@
...
+// initial delay and a factor applied to it for each additional failed attempt.
+//
+// If the number of failures reaches a maximum, the IP address is blocked for a
+// certain amount of time. Further attempts after the end of this reset duration
+// are allowed and the first one resets the counter, allowing failures with
+// short delay again. All reservations during the reset duration are denied.
+type IPLoginDelayer struct {
+	config   LoginDelayerConfig
+	delayers map[string]*loginDelayer
+	mutex    *sync.Mutex
```
krilor, line 32: NIT: I don't know but I think it is a bit uncommon to store a pointer to a mutex like this. At least had to stop and think when I saw it. Having the actual mutex on the struct makes more sense IMHO

backend/auth/iplogindelayer.go
```diff
@@ -0,0 +1,166 @@
...
+
+type loginDelayer struct {
+	ipAddress            string
+	config               *LoginDelayerConfig
+	mutex                *sync.Mutex
+	currentFailedLogins  int
+	lastLoginAttemptTime time.Time
+}
+
+func (delayer *loginDelayer) Allow(ctx context.Context) (bool, *time.Time) {
```
krilor, line 88: NIT: I would expect this signature to just return bool. Especially since there is no comment on this method to tell the user what the time.Time is. Also, most of the time it will just be true. Getting the time could/should be just another method, e.g. RetryAfter that returns a time in UTC that can be returned to client with http.TimeFormat

backend/auth/iplogindelayer.go
```diff
@@ -0,0 +1,166 @@
...
+		delayer.currentFailedLogins = 0
+	}
+
+	// user reached the maximum amount of failed logins
+	if delayer.currentFailedLogins >= delayer.config.MaxFailedLogins {
+		slog.DebugContext(
+			ctx, "IP has reached the maximum number of failed logins",
+			"ipAddress", delayer.ipAddress,
+		)
+		return false, &resetTime
```
krilor, line 122: Why are we returning the resetTime here? It feels a bit off that this function either returns delayTime or resetTime, and the client does not know what it gets. Something feels a bit off here.

backend/auth/iplogindelayer.go
```diff
@@ -0,0 +1,166 @@
...
+	mutex                *sync.Mutex
+	currentFailedLogins  int
+	lastLoginAttemptTime time.Time
+}
+
+func (delayer *loginDelayer) Allow(ctx context.Context) (bool, *time.Time) {
+	delayer.mutex.Lock()
+	defer delayer.mutex.Unlock()
+
+	delayedTime := delayer.lastLoginAttemptTime.Add(delayer.computeDelay())
```
krilor, line 92: NIT: Tiny optimization and possibly making it a bit easier to reason about things: store the time of when the user can retry after rather than when they tried? So instead of storing lastLoginAttemptTime, store retryAfter.

krilor (changes requested):

backend/auth/iplogindelayer.go
enzo-crance-statnett: I'm trying to be careful in this territory, cf [official docs](https://pkg.go.dev/sync#Mutex)
> _A Mutex must not be copied after first use._

Pointer is safest, even though it is a bit heavier.

But as there is only one `IPLoginDelayer` (it is never copied) and the underlying `loginDelayer` values are stored as pointers in the map, it should be fine if we remove pointers in front of mutexes.

backend/auth/iplogindelayer.go
enzo-crance-statnett: In such a case, we would just be able to know if the request is allowed or not, IIUC we would lose the ability to differentiate between delay and reset time, so we could not tell whether to reset the login counter.

krilor (dismissed):

PR #164, opened 2025-05-09
chore: pin PostgREST version before v13 (FLEX-583)

enzo-crance-statnett:
Since the latest major change introduced breaking changes for us, we first need to pin the version to the previous one so the system continues to work.


krilor (dismissed):
Should have a TODO comment in the file for why this is there.

PR #167, opened 2025-05-09
feat(ui): show list of BRPs associated to each controllable unit (FLEX-576)

enzo-crance-statnett:
This PR adds a sublist of BRPs to the controllable unit show page.

### DIFF
--- a/frontend/src/controllable_unit/ControllableUnitShow.tsx
+++ b/frontend/src/controllable_unit/ControllableUnitShow.tsx
@@ -19,6 +19,7 @@ import { TechnicalResourceList } from "./technical_resource/TechnicalResourceLis
 import { UnitField } from "../unitComponents";
 import { EventButton } from "../event/EventButton";
 import { IdentityField } from "../IdentityField";
+import { ControllableUnitBalanceResponsiblePartyList } from "./balance_responsible_party/ControllableUnitBalanceResponsiblePartyList";
 
 export const ControllableUnitShow = () => {
   const resource = useResourceContext()!;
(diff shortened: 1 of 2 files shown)


enzo-crance-statnett:
Draft because I haven't managed to synchronise both queries yet.

The one on the AP resource and the one on the AP-BRP resource. Currently both queries are run at the same time which means the first call to AP-BRP does not filter, so we get all the visible BRPs we can see regardless of the AP. Big performance issue there.

The second query is then re-run with the right filter, but we still get this weird thing that does not get updated in the pagination. cf screenshot showing 12 and a second page instead of just 4. If we click on the second page we get an error.

Screenshot 2025-05-09 164653

WIP

enzo-crance-statnett:
Here is the flow of information:

- the component is shown on the CU page, so the record gives us the `CU.accounting_point_id` (corresponds to `AP.business_id`)
- we call the API to get `AP.id` from `AP.business_id`
- we call the API to get the BRPs readable on this AP.

The second step there will be annoying everytime we want to show something based on AP on the CU page. However, the relations we get from the external data are based on AP, so it makes sense to keep AP-BRP and AP-ES in the database, not do some kind of CU-BRP for instance. But in the current state, if all the queries related to BRP/ES start from a CU, then the API users will always have to do this translation step.

I think the fundamental question there is whether to keep an AP _business ID_ on the CU resource. All the `AP-*` resources expect a technical ID, and the AP resource is open (`AP-COM001`) so the API users can always get the associated business ID if needed. IMO it wouldn't hurt to switch to `CU.accounting_point_id` being a foreign key to `AP.id` instead of a business ID, and use a `ReferenceField` in the UI everywhere there is an AP to show, so the UI users always see the business ID. It is indeed a breaking change but I think we talked about it earlier several times.

> [!NOTE]
> From a pure frontend point of view, in any case, there should be a way to synchronise queries. It's bad to have to first query the full list, then re-run the query with a filter. But here we have an opportunity to remove the problem instead of having to fix it.

enzo-crance-statnett:
Now the frontend does no longer ask for the whole AP-BRP resource, and pagination works. Still draft because it is now dependent on https://github.com/elhub/flex-information-system/pull/170.

PR #160, opened 2025-05-08
chore: cleanup post liquibase (FLEX-502)

krilor:

PR #170, opened 2025-05-12
feat: turn controllable_unit.accounting_point_id into a technical ID (FLEX-576)

enzo-crance-statnett:


db/flex/controllable_unit.sql
```diff
@@ -58,9 +56,10 @@ CREATE TABLE IF NOT EXISTS controllable_unit (
         localtimestamp, NULL, '[)'
     ),
     recorded_by bigint NOT NULL DEFAULT current_identity(),
+
     CONSTRAINT controllable_unit_accounting_point_id_fkey FOREIGN KEY (
         accounting_point_id
-    ) REFERENCES accounting_point (business_id),
+    ) REFERENCES accounting_point (id),
```
krilor, line 62: Now that we are using liquibase and actually doing database migrations instead of recreating the DB all-the-time we need to write a migration for this.

As we landed in https://github.com/elhub/flex-information-system/pull/160/commits/50fabd763ff879a072841f4d814ac88c75642f42 we still want to update this table definition, but _in addition_ we must write migrations, aka ALTER TABLE, DROP/CREATE CONSTRAINT etc

The test is

* "just reload" on main
* "just liquibase" on this branch
* verify that the table looks the way it is supposed to

Note that the migrations probably also should have preconditions to prevent them from running if the table definition is already correct (e.g. for new deployments/environments)

We could also look at tools such as https://github.com/djrobstep/migra to help with the migration authoring. I'm suggesting migra since the supabase people are betting on it for their schema tooling

krilor (changes requested):
Looking good. Need to handle the DB bits as noted in an inline comment.

I'm also a bit worried that we are not touching the ControllableUnitInput and ControlableUnitLookup UI pages. I would assume that this change requires some changes there.

db/flex/controllable_unit.sql
enzo-crance-statnett: Added the migration and docs for future ones too, based on this comment and our discussion.

docs-dev/db-migration.md
krilor, line 1: This is good stuff, but overlapping with https://github.com/elhub/flex-information-system/blob/main/docs-dev/db-design.md#schema-maintenance-and-migrations

Can you please merge the stuff from db-design.md into here and possibly change the page title to the one used in db-design.md?

docs-dev/editing-resources.md
```diff
@@ -63,6 +63,9 @@ operations:
 If an operation should not be allowed, just remove it from this list and the
 associated endpoints will not be generated in the OpenAPI documentation.
 
+> [!WARNING]
```
krilor, line 66: NIT: Should use link?

frontend/src/controllable_unit/lookup/ControllableUnitLookupResult.tsx
```diff
@@ -75,7 +76,12 @@ const ControllableUnitLookupResultItem = () => {
         >
           <TextField source="id" label="ID" />
           <TextField source="business_id" label="Business ID" />
-          <TextField source="accounting_point_id" label="Accounting point ID" />
+          <ReferenceField
```
krilor, line 79: Should we consider showing both the technical id and the GSRN here?

krilor (dismissed):
I added a few inline comments that should be adressed/responded to, but this is looking goood 👏

PR #172, opened 2025-05-15
feat(backend): explicit HTTP routes (FLEX-524)

enzo-crance-statnett:
This PR makes all HTTP routes explicit, so that we avoid bulk updates/deletes for instance, or subpaths that could work in PostgREST but that we did not plan to open.


krilor (approved):
Looking good! But we should take the oppurtunity to delete the "subresource" code that is now dead ( https://github.com/elhub/flex-information-system/blob/main/backend/data/data.go#L290-L312 ) and related docs ( https://github.com/elhub/flex-information-system/blob/main/docs/api-design.md#hierarchies-via-redirects ) We can do that in this PR or as a follow-up.

PR #174, opened 2025-05-16
feat: update product types (FLEX-537)

enzo-crance-statnett:
This PR introduces a migration to apply the changes to product types listed in https://github.com/elhub/flex-information-system/commit/3ed21aaa1dbcb96e8dedce4e7518328071cbdbab.


db/test_data/test_data.sql
```diff
@@ -615,9 +615,10 @@ BEGIN
   INSERT INTO flex.system_operator_product_type (
     system_operator_id,
     product_type_id
-  ) SELECT so_id, id
-  FROM flex.product_type pt
-  WHERE pt.business_id in ('manual_congestion_activation', 'manual_congestion_capacity');
+  ) VALUES (
+    so_id,
+    1
```
krilor, line 620: Should we really do this hardcoding of id?

db/flex/product_type_migrations.sql
```diff
@@ -0,0 +1,201 @@
...
+
+-- changeset flex:product-type-list-update runOnChange:false endDelimiter:;
+--preconditions onFail:MARK_RAN
+--precondition-sql-check expectedResult:1 SELECT COUNT(*) FROM information_schema.columns WHERE table_schema = 'flex' AND table_name = 'product_type' AND column_name = 'market'
+
+SELECT set_config('flex.current_identity', '0', false);
+
+-- new columns
+ALTER TABLE flex.product_type ADD COLUMN name text;
+ALTER TABLE flex.product_type ADD COLUMN service text;
```
krilor, line 12: NIT: I agree that we added "name", but the other two are basically renames

* category -> service
* example -> products

db/flex/reference_data/product_type.csv
```diff
@@ -1,12 +1,10 @@
...
+BUSINESS_ID,NAME,SERVICE,PRODUCTS
+manual_congestion,Manual Congestion,congestion management,"LongFlex, ShortFlex"
+automatic_congestion,Automatic Congestion,congestion management,MaxUsage
+automatic_voltage,Automatic Voltage,voltage control,
+manual_voltage,Manual Voltage,voltage control,
+automatic_fast_frequency,Fast Frequency,balancing,"FFR Profil, FFR Flex"
+automatic_frequency_containment,Frequency Containment,balancing,"FCR-N, FCR-D"
+automatic_frequency_restoration,Automatic Frequency Restoration,balancing,aFRR CM
+manual_frequency_restoration,Manual Frequency Restoration,"balancing","mFRR EAM, mFRR CM"
+manual_frequency_restoration_disruption,Frequency Disruption,"balancing","mFRR-D"
```
krilor, line 10: We should probably name this last one "Frequency Restoration Disruption"

krilor (changes requested):
Good work. Requesting changes for the non-NIT/TIP comments.

db/flex/product_type_migrations.sql
```diff
@@ -0,0 +1,201 @@
...
+ALTER TABLE flex.product_type ADD COLUMN products text;
+
+-- avoid having to move IDs when possible, by reusing existing entries
+UPDATE flex.product_type
+SET
+    business_id = 'manual_congestion',
+    name = 'Manual Congestion',
```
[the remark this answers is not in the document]
enzo-crance-statnett: TIL

Still filling `name` though, because if we want it to be non-nullable we must have a value there for all entries.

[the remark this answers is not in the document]
krilor: Yeah, I get that. Another strategy would be to do something like this early in the changeset. This also allows us to keep the column definitions close together - which is easier to follow as a reader.

```sql
ALTER TABLE flex.product_type ADD COLUMN name text;
UPDATE flex.product_type SET name = business_id;
ALTER TABLE flex.product_type ALTER COLUMN name SET NOT NULL;
```

krilor (approved):

PR #177, opened 2025-05-20
feat(backend): better logging for token decoding (FLEX-604)

enzo-crance-statnett:
First step to make things clearer.

### DIFF
--- a/backend/auth/auth.go
+++ b/backend/auth/auth.go
@@ -106,7 +106,11 @@ func NewAPI(
 	}
 }
 
+const tokenDecodingLogContext = "token decoding"
+
 // TokenDecodingMiddleware decodes the token and sets it as RequestDetails in the context.
+//
+//nolint:funlen
 func (auth *API) TokenDecodingMiddleware(
 	next http.Handler,
 ) http.Handler {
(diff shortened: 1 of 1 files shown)


backend/auth/auth.go
```diff
@@ -143,8 +151,16 @@ func (auth *API) TokenDecodingMiddleware(
 				})
 				return
 			}
+			slog.InfoContext(
+				ctx, "found token in auth header",
```
krilor, line 155: Consider leaving this for Debug. Having this for all requests seems unneccessary.

backend/auth/auth.go
```diff
@@ -106,7 +106,11 @@ func NewAPI(
 	}
 }
 
+const tokenDecodingLogContext = "token decoding"
```
krilor, line 109: Do we really need this context bit? We've been discussing this before, and the only time we need it is in debug, and there we have replaced it with addSource ref https://github.com/elhub/flex-information-system/blob/main/backend/flex.go#L117

The messages are ok on their own imho

krilor (approved):
Seems a bit verbose. Consider fixing that. See inline comments.

PR #190, opened 2025-05-26
feat(db): notice SP when one of their SPG has multiple BRPs (FLEX-489)

enzo-crance-statnett:
Based on https://github.com/elhub/flex-information-system/pull/142, adapted to later changes and fixed.


krilor (changes requested):
We must put this notice in our list

https://github.com/elhub/flex-information-system/blob/main/docs/resources/notice.md

krilor (dismissed):

PR #193, opened 2025-05-28
feat: remove non-stable IDs from notice and AP-BRP (FLEX-601)

enzo-crance-statnett:
This PR removes non-stable IDs from notice and AP-BRP resources.

BRPs on a CU:

Screenshot 2025-05-28 155407

Notices:

Screenshot 2025-05-28 161511


db/flex/notice.sql
enzo-crance-statnett, line 1: Apart from removing the call to `row_number` for the non-stable ID, only indentation was changed in this file.

krilor (approved):
Sweet!

PR #198, opened 2025-06-02
feat: accounting point energy supplier resource on the API (FLEX-571)

enzo-crance-statnett:
This PR opens AP-ES to the API. It is shown in the UI as a part of the CU page, very much like AP-BRP.

Regarding RLS, this PR also follows what had been done with AP-BRP. We use the same policies, and as these policies are not only restricting or allowing access but also modifying the data based on the role, they are directly implemented in the view.

This resource has no ID, for uniformity with AP-BRP, and because it is useless as there is no show/edit page and no single-row operation defined in the API.

### DIFF
--- a/docs/resources/accounting_point_energy_supplier.md
+++ b/docs/resources/accounting_point_energy_supplier.md
@@ -75,9 +75,7 @@ No policies.
 
 #### Service Provider
 
-| Policy key | Policy                                                         | Status |
-|------------|----------------------------------------------------------------|--------|
-| APES-SP001 | Read APES on periods where they are related to a CU on the AP. | DONE   |
+No policies.
 
 #### Third Party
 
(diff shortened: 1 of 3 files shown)


docs/resources/accounting_point_energy_supplier.md
```diff
@@ -0,0 +1,95 @@
...
+
+| Policy key | Policy                                             | Status |
+|------------|----------------------------------------------------|--------|
+| APES-SO001 | Read APES when they are system operator on the AP. | DONE   |
+
+#### Service Provider
+
+| Policy key | Policy                                                         | Status |
+|------------|----------------------------------------------------------------|--------|
+| APES-SP001 | Read APES on periods where they are related to a CU on the AP. | DONE   |
```
krilor, line 80: We should consider if SP should have read or not. Generally, SP does not require this information for their operation, and one could say that it is "market information" that we should not expose.

krilor (approved):

PR #202, opened 2025-06-04
docs: recipe to deploy mkdocs to GitHub Pages (FLEX-620)

enzo-crance-statnett:


krilor (dismissed):
Good stuff. Lets include a step to check for main as we talked about.

PR #206, opened 2025-06-05
feat: deploy API docs also in GH Pages + small fixes (FLEX-620)

enzo-crance-statnett:


local/elements/index.html
```diff
@@ -34,12 +33,7 @@
     />
     <footer id="link-footer">
       <form style="display: flex; gap: 10px">
-        <a class="link-footer-button" href="/">Portal</a>
-        <a
-          class="link-footer-button"
-          href="https://elhub.github.io/flex-information-system"
-          >Project documentation</a
-        >
+        <a class="link-footer-button" href="/">← Back</a>
```
enzo-crance-statnett, line 36: This back button allows not having to hardcode the portal's address (because if Elements docs are deployed to GitHub Pages then `/` does no longer point to the portal).

local/scripts/build-static.sh
```diff
@@ -58,6 +58,10 @@ function openapi() {
 		'.servers = [$servers.'"$api_name.$env_short"']' <"openapi/openapi-${api_name}.json" \
 		>"./dist/$api_name/v0/openapi.json"
 
+	mkdir -p "./docs/$api_name/v0"
+	cp "./dist/$api_name/v0/openapi.json" "./docs/$api_name/v0"
```
krilor, line 62: It seems off that docs publishing should be dependent on this script being run. "just docs" should work in isolation and implicitly require this to have been run seems like a footgun to me.

docs/resources/accounting_point.md
enzo-crance-statnett, line 1: Because of the internal tag `#/operations/...`, mkdocs does not recognise the relative link as valid and somehow generates a wrong link with still `/resources` in it, and for the same reason as in the other comment, using an absolute link wouldn't work both on local and on the docs website. Fixed by just always pointing to the docs website.

PR #208, opened 2025-06-05
feat(backend): add default limit parameter on list endpoints (FLEX-596)

enzo-crance-statnett:
Also added the list query parameters to history resources in the API. Default limit set to 50.


backend/internal/middleware/limit.go
```diff
@@ -0,0 +1,26 @@
...
+	"regexp"
+	"strconv"
+)
+
+const defaultLimit = 50
+
+// Limit adds a default limit query parameter to requests on list endpoints if
+// not provided by the client, so that we never query a whole resource at once.
+func Limit(next http.Handler) http.Handler {
+	regexSingleID := regexp.MustCompile("/[0-9]+$")
```
krilor, line 14: Consider just adding the middleware in the places we want it rather than doing this pattern matching? We already have explicit routes configured for list operations (GET collection)

krilor (approved):

PR #215, opened 2025-06-13
feat(db): landing table for external party updates (FLEX-631)

enzo-crance-statnett:


krilor (changes requested):
* add ID
* unique constraint on type + gln
* move the test stuff to test_data

db/test_data/test_data.sql
```diff
@@ -839,3 +839,23 @@ BEGIN
...
+  SELECT
+      p.business_id AS party_gln,
+      e.business_id AS entity_org,
+      p.name,
+      p.type
+  FROM flex.party AS p
+      INNER JOIN flex.entity AS e ON p.entity_id = e.id
+  WHERE p.business_id_type = 'gln'
+      AND e.business_id_type = 'org'
+      AND p.status != 'terminated';
```
krilor, line 859: NIT: On conflict do nothing?

db/flex/party_staging.sql
```diff
@@ -0,0 +1,26 @@
+--liquibase formatted sql
+-- Manually managed file
+
+-- landing table for party data updates coming from external sources
+-- NOTE: when testing locally, run test_data.fill_party_staging() to populate
+-- changeset flex:party-staging-create runOnChange:false endDelimiter:--
+CREATE TABLE IF NOT EXISTS party_staging (
+    id bigint PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+    party_gln text CHECK (validate_business_id(party_gln, 'gln')),
```
krilor, line 9: NIT: I'd remove the `party_` and `entity_` bit here. I get where it is coming from, and they kindof make sense. Reasons:

* `party.party_` does not sit well with me. Its just uneccessary stutter.
* What about "name" and "type"? If we do it for gln, then we kind of need to do it for those as well?

krilor (approved):

PR #216, opened 2025-06-16
feat: notice FISOs on diff between party_staging and party (FLEX-632)

enzo-crance-statnett:
Screenshot 2025-06-16 141758

> [!NOTE]
> The second notice, tracking updates, currently assumes as a first approximation that changes on the party resource can happen on `name` and `entity_id`, because GLN is stable and I assumed `type` was too. This can be refined if needed.


db/flex/notice.sql
```diff
@@ -152,4 +152,70 @@ CREATE OR REPLACE VIEW notice AS (
...
+        )
+
+    -- parties already in the system, but with staging updates
+    UNION ALL
+    SELECT
+        p_fiso.id AS party_id,
+        'no.elhub.flex.party.updated' AS type, -- noqa
+        '/party/' || p.id AS source,
+        jsonb_build_object(
+            'name', p_stg.name,
```
krilor, line 188: Is there a way to only include a field if it is actually updated/different?

db/flex/notice.sql
```diff
@@ -152,4 +152,70 @@ CREATE OR REPLACE VIEW notice AS (
...
+        p_fiso.id AS party_id,
+        'no.elhub.flex.party.new' AS type, -- noqa
+        null AS source, -- no source because the party does not exist yet
+        jsonb_build_object(
+            'business_id', p_stg.gln,
+            'business_id_type', 'gln',
+            'entity_id', e_stg.id,
+            'name', p_stg.name,
+            'type', p_stg.type,
+            'role', 'flex_' || p_stg.type
```
krilor, line 168: "role" is very much "internal" and probably should not be exposed (?)

db/flex/notice.sql
```diff
@@ -152,4 +152,70 @@ CREATE OR REPLACE VIEW notice AS (
...
+        )::text AS data -- noqa
+    FROM flex.party_staging AS p_stg -- noqa
+        INNER JOIN flex.entity AS e_stg
```
krilor, line 171: In the flow of populating the register, could it not be that the entity does not in the register? This means that an inner join here is a bit scary?

db/flex/notice.sql
```diff
@@ -152,4 +152,70 @@ CREATE OR REPLACE VIEW notice AS (
...
+            ON p_stg.org = e_stg.business_id
+        -- warn all FISOs
+        INNER JOIN flex.party AS p_fiso
+            ON p_fiso.type = 'flexibility_information_system_operator'
+
+    -- parties deleted after sync, but still actually in the system
+    UNION ALL
+    SELECT
+        p_fiso.id AS party_id,
+        'no.elhub.flex.party.deleted' AS type, -- noqa
```
krilor, line 208: Is "deleted" and "new" good words for these _notices_?  It smells like events. For new, maybe "missing" is better? For "deleted", I'm not sure... reduntant? extra? not_valid? ???

docs/resources/notice.md
```diff
@@ -7,17 +7,20 @@ Notice to users about various issues or actions expected from them.
...
+| no.elhub.flex.service_providing_group_product_application.status.requested    | SPG product application status requested                             | PSO               | Initiate SPG product prequalification and update status                                           |
+| no.elhub.flex.service_providing_group_grid_prequalification.status.requested  | SPG grid prequalification status requested                           | ISO               | Initiate SPG grid prequalification and update status                                              |
+| no.elhub.flex.service_providing_group.balance_responsible_party.multiple      | Inconsistency: Multiple BRPs in a single SPG                         | SP                | Make sure the SPG only contains CU currently associated to the same BRP on their accounting point |
+| no.elhub.flex.controllable_unit_service_provider.valid_time.outside_contract  | Inconsistency: CUSP valid while end user is not valid on the AP      | SP                | Update CUSP to match the updated end user data from Elhub                                         |
```
krilor, line 20: Should we sort this list?

docs/resources/notice.md
```diff
@@ -7,17 +7,20 @@ Notice to users about various issues or actions expected from them.
...
+| no.elhub.flex.party.missing                                                   | New party fetched from an external source but not registered         | FISO              | Create party to match the latest data fetched from the external source                            |
+| no.elhub.flex.party.outdated                                                  | Party not in the latest state fetched from an external source        | FISO              | Update party to match the latest data fetched from the external source                            |
+| no.elhub.flex.party.residual                                                  | Party still in the system but no longer listed by an external source | FISO              | Delete party to match the latest data fetched from the external source                            |
```
krilor, line 23: NIT:
"New party fetched..."
"Party still in..."

This assumes that events/things happen in a certain order. I can think of some (at least error) scenarios where these texts are not true. Its might be better to state that the party is/not present in the authorative source at the current time.

krilor (approved):

PR #218, opened 2025-06-17
feat(ui): buttons from party notices to autofilled party input page (FLEX-633)

enzo-crance-statnett:
Screenshot 2025-06-17 140940

> [!NOTE]
> Null `entity_id` comes from the fact the org number mentioned in the `party_staging` record does not match an existing entity. We allowed this to happen in the previous PR.


enzo-crance-statnett:
> [!NOTE]
> Null `entity_id` comes from the fact the org number mentioned in the `party_staging` record does not match an existing entity. We allowed this to happen in the previous PR.
>
> It is possible to show the org number instead of the entity ID in these cases though, if it makes sense. We just have to name it differently from `entity_id` and it will be filtered out of the record by the support buttons added in this PR, so normally it won't break anything.

enzo-crance-statnett:
As discussed, removing the termination button.

PR #219, opened 2025-06-17
feat: enable assume party for entities on owned parties (FLEX-636)

enzo-crance-statnett:


docs/resources/party_membership.md
```diff
@@ -35,9 +35,9 @@ resource. The default policy is **deny**.
 
 #### Entity
 
-| Policy key  | Policy                                          | Status |
-|-------------|-------------------------------------------------|--------|
-| PTYM-ENT001 | Read all the PTYM concerning the current entity | DONE   |
+| Policy key  | Policy                                     | Status |
+|-------------|--------------------------------------------|--------|
+| PTYM-ENT001 | Read all the PTYM on parties they can read | DONE   |
```
krilor, line 40: We cannot have this policy. Parties (!= end_user) is _generally_ open information. The fact that we currently have it "closed" for flex_entity is not something we can rely on staying that way. At some point, we might even consider anonomous access to that information.

Why are we changing it like this?

db/flex/party_rls.sql
```diff
@@ -22,9 +22,29 @@ GRANT SELECT ON party TO flex_entity;
...
 TO flex_entity
-USING (
-    EXISTS (SELECT 1 FROM party_membership pm WHERE pm.party_id = party.id) -- noqa
-);
+USING (party.entity_id = flex.current_entity());
+
+-- RLS: PTY-ENT002
+-- security definer function avoids round trips in RLS policies
+-- (PTYM-ENT001 checks the party table)
+CREATE FUNCTION is_entity_in_party(
```
krilor, line 30: NIT: `is_entity_of_party` ?

docs/resources/party.md
```diff
@@ -60,9 +60,10 @@ resource. The default policy is **deny**.
 
 #### Entity
 
-| Policy key | Policy                                       | Status |
-|------------|----------------------------------------------|--------|
-| PTY-ENT001 | Read all PTY whose memberships they can read | DONE   |
```
krilor, line 65: I liked this policy and think it should stay that way. Why are we changing it?

enzo-crance-statnett:
We are in a situation where the implementation of RLS makes one of the resources depend on the other. This PR introduces a round trip and we have to be careful when we implement it (especially when there exist several ways). The current one is just the first one for which I became confident enough it would work.

In the current state, as `flex_entity` is not a `flex_common`, nothing has changed radically. But I missed that piece of information about the party resource potentially becoming more open in the future. That makes the RLS options not all equally acceptable, and I agree that we should not make the dependency in that direction, so that we can open party without breaking RLS for memberships later on.

The following implementation should fix it in the right direction:

- entity sees parties when they can see a membership
- entity can see membership when they are the entity
- entity can see membership when they own the party (security definer check)

krilor (dismissed):

PR #225, opened 2025-06-19
feat: make notice.data an object, specified in OpenAPI docs (FLEX-634)

enzo-crance-statnett:
This PR turns the `data` field of the notice resource from a string into a specified object with several possible schemas based on the notice type.

For best OpenAPI support and least complexity, we keep the specification quite simple with a simple `oneOf` between the schemas without further dependencies on the `type` field. Some docs are added for the notice resource to compensate and help API users understand what they get.

Screenshot 2025-06-19 133210

Screenshot 2025-06-19 133218

Screenshot 2025-06-19 133225

Updated changelog as this is a "semi-breaking" change (the type change is breaking but in a way we are just making a field _more specified_ here).

### DIFF
--- a/backend/data/models/models.go
+++ b/backend/data/models/models.go
@@ -145,7 +145,7 @@ type Notice struct {
 	PartyID int
 	Type    string
 	Source  string
-	Data    *string
+	Data    []byte
 }
 
 type Notification struct {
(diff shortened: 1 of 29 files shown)


enzo-crance-statnett:
As we have discussed the design more thoroughly and made significant changes, I am closing the current PR in favour of https://github.com/elhub/flex-information-system/pull/230

PR #230, opened 2025-06-20
feat: turn notice.data field into a specified object

enzo-crance-statnett:


enzo-crance-statnett:
UI changes are limited to compatibility for now (fixing the current buttons so they continue to work).

enzo-crance-statnett:
Some examples

Screenshot 2025-06-20 140512

Screenshot 2025-06-20 140538

backend/data/static/openapi.json
```diff
@@ -255,6 +255,70 @@
...
+                        "to": {
+                            "type": "string",
+                            "format": "date-time",
+                            "description": "The end time of the timeline section.",
+                            "example": "2023-01-02T00:00:00Z"
+                        }
+                    }
+                }
+            },
+            "notice_data_cusp_valid_time_outside_contract": {
```
krilor, line 279: Any reason for tagging this specifically for CUSP? When we talked this morning, we talked about a reusable schema what we could have for both CUSP and SPGM outside_contract?

backend/data/static/openapi.json
```diff
@@ -255,6 +255,70 @@
                     "technical_resources"
                 ]
             },
+            "timeline_data": {
+                "description": "Partial timeline data that is relevant to mention, in a notice for instance. Multirange format, i.e., array of start/end timestamp pairs.",
+                "type": "array",
+                "items": {
+                    "type": "object",
+                    "properties": {
+                        "from": {
```
krilor, line 264: Shouldn't we use use the naming that we have for timeline? valid_from and valid_to?

backend/data/static/openapi.json
```diff
@@ -255,6 +255,70 @@
                     "technical_resources"
                 ]
             },
+            "timeline_data": {
```
krilor, line 258: `timeline_multi_range` ?

local/scripts/template_to_openapi.py
```diff
@@ -559,6 +559,20 @@ def generate_openapi_document(base_file, resources_file, servers_file):
                 "type": "object",
             }
 
+        # add rawdata schema (all nullable, nothing required) for inclusion/embedding in other
```
krilor, line 562: Not sure that "raw data" conveys the meaning that we want? Is it really "raw data" ?

openapi/openapi-api-no-default.json
```diff
@@ -998,6 +1062,47 @@
                 ],
                 "type": "object"
             },
+            "entity_rawdata": {
+                "summary": "Raw data - Entity",
+                "description": "Raw data for embedding - Entity - Natural or legal person\n\nAn entity is a natural or legal person that can be a party in the Flexibility Information System.\n\nExample entity types:\n\n* Person\n* Organisation",
```
krilor, line 1067: We had a lot of discussions and thinking to ensure that we only define the fields once in in the spec, with inheritance like: `_response` inherits `_create_request`, inherits `_update_request`.

Now we are just ignoring that part and going "raw". Should we really? Or should we continue with this inheritance scheme?

We discussed this morning that we do _not_ want the "null" to pop out into the notice.data field (we rather remove those). The only other need we have is to remove the "required" part. Should be doable.

db/flex/notice.sql
```diff
@@ -132,8 +134,23 @@ CREATE OR REPLACE VIEW notice AS (
...
         jsonb_build_object(
             'invalid_timeline',
-            valid_time_range - end_user_timeline
-        )::text AS data -- noqa
+            (
+                SELECT
+                    array_to_json(
+                        array_agg(
+                            jsonb_build_object(
+                                'from', lower(timeline_section),
```
krilor, line 142: valid_from|to

db/flex/notice.sql
```diff
@@ -160,12 +177,18 @@ CREATE OR REPLACE VIEW notice AS (
...
         null AS source, -- no source because the party does not exist yet
         jsonb_build_object(
-            'business_id', p_stg.gln,
-            'business_id_type', 'gln',
-            'entity_id', e_stg.id,
-            'name', p_stg.name,
-            'type', p_stg.type
-        )::text AS data -- noqa
+            'entity', jsonb_build_object(
+                'business_id', p_stg.org,
```
krilor, line 181: missing business_id_type and type?

db/flex/notice.sql
```diff
@@ -160,12 +177,18 @@ CREATE OR REPLACE VIEW notice AS (
...
+                'name', e_stg.name
+            ),
+            'party', jsonb_build_object(
```
krilor, line 184: jsonb_strip_nulls ?

db/flex/notice.sql
```diff
@@ -183,24 +206,22 @@ CREATE OR REPLACE VIEW notice AS (
...
-                    THEN jsonb_build_object('name', p_stg.name)
-                ELSE '{}'::jsonb
-            END
-        ) || (
-            CASE WHEN p_stg.org != e.business_id
-                    THEN jsonb_build_object('entity_id', e_stg.id)
-                ELSE '{}'::jsonb
-            END
-        ))::text AS data -- noqa
+        CASE WHEN p_stg.name != p.name
```
krilor, line 209: I think the CASE is not needed anymore since we dropped the org bits. Same with join to entity.

But why are we dropping the org check?

backend/data/static/openapi.json
enzo-crance-statnett: The reason is: the SPGM notice does not have any data ATM.

We could compute the invalid part of the timeline in this other notice type as well, but currently we do not do it. Changing the spec and associating the data format to the notice would require to write that computation. Maybe it's the right thing to do.

But the CUSP points to the `timeline_multi_range` precisely to be able to reuse that schema elsewhere. So we have the tools ready for more notices of that kind. We actually are generic/reusable already.

local/scripts/template_to_openapi.py
enzo-crance-statnett: Renamed to "optional" to be more explicit.

openapi/openapi-api-no-default.json
enzo-crance-statnett: I removed the nullable markers. Indeed we just talked about non-required, not actually nullable.

Regarding the schema references. We could have reused response schemas indeed. I have tried though and haven't managed to override required fields nested under schema references. This is why I went with redeclaring a non-required version of the schema.

A third way would be to change the way we mark stuff as required, for instance having no required fields in the schemas, but having the `required` list at the place they are referenced, _i.e._, in each of the endpoint specs.

db/flex/notice.sql
enzo-crance-statnett: Re-added both as we agreed entity can change. We now have an empty object though, this does not break the spec per se but it could be made cleaner.

Screenshot 2025-06-23 161104

(first line here is a name update, second line is an entity update with an entity that doesn't exist yet)

backend/data/static/openapi.json
```diff
@@ -1125,6 +1243,79 @@
                     }
                 }
             },
+            "party_create_data": {
```
krilor, line 1246: I was expecting to see an update to party_create_request to be something like

```json
...
"party_create_request": {
    "summary": "Create - Party",
    "description": "Request schema for create operations - The body that interacts with the Flexibility Information System\n\nA party is the thing that is authorized to access or modify data in the Flexiblity Information System.\n\nExample party types:\n\n* Service Provider\n* System Operator\n* End User",
    "allOf": [
      {
        "$ref": "#/components/schemas/party_create_data"
      },
      {
        "required": [
          "business_id_type",
          "entity_id",
          "name",
          "role",
          "type",
          "status"
        ]
      }
...
```

krilor: The schemas  `party` and `party_response` is also duplicating fields. I thought we agreed to avoid that ?

docs-dev/diagrams/openapi-schemas-base.png
krilor, line 1: These diagrams does not follow our naming standard https://github.com/elhub/flex-information-system/blob/main/docs-dev/principles.md#diagrams . 

Expecting that this is a drawio diagram and should then be `.drawio.png` . If it is not a drawio diagram then we need to document the new tool and make sure that it is editiable by contributors.

docs-dev/openapi-schemas.md
```diff
@@ -0,0 +1,87 @@
...
+The smallest schema is the **update** schema, that contains all properties
+_except_ the _non-updatable_ and _read-only_ ones (so basically all the
+properties that can change over time).
+
+The **create** schema is obtained by adding the _non-updatable_ fields and
+marking the _required_ fields based on our definition above.
+
+Finally, the **response** schema is obtained by adding the _read-only_ fields.
+
+Sometimes, there is no create operation for the resource.
```
krilor, line 52: This is the case also for update

Like this empty one in our current output.

```
"party_membership_update_request": {
    "summary": "Update - Party Membership",
    "description": "Request schema for update operations - The relation between a party and entity.",
    "type": "object"
  },
```

docs-dev/openapi-schemas.md
```diff
@@ -0,0 +1,87 @@
...
+In such cases, it makes no sense to create a **create** schema, because it would
+just be confusing to read for the user.
+Therefore, we adopt the following, simplified generation strategy:
+
+![Structure without create operation](./diagrams/openapi-schemas-no-create.png)
+
+## Data schemas
```
krilor, line 59: This doc seems to state that we want to treat some resources in a special way. Why do we want that? Cant we just do the same for all? Does the "data" schema hurt for other resources? Cant we just generate all for all resources? Wouldnt that be the least complex thing to do?

docs-dev/openapi-schemas.md
krilor, line 1: This is good stuff, but the thing I feel like we are missing are the actual schema names.
Consider turning the text into a table that holds

* name - like `<resource>_create_request`
* embeds - like `<resource>_create_data` and `requires`
* description - some of the text
* description

db/flex/notice.sql
```diff
@@ -183,26 +212,36 @@ CREATE OR REPLACE VIEW notice AS (
...
-                    THEN jsonb_build_object('entity_id', e_stg.id)
-                ELSE '{}'::jsonb
-            END
-        ))::text AS data -- noqa
+        jsonb_strip_nulls(
+            jsonb_build_object(
+                'entity',
+                (CASE WHEN p_stg.org != e.business_id
+                        THEN jsonb_build_object(
+                                'business_id', p_stg.org,
```
krilor, line 220: Are we missing `id` here?

enzo-crance-statnett:
Went more agressively into the merging strategy.

- generate the hierarchy all the time, even when there is no create
- `create_data` is the only schema with required annotations. It is re-used in both `create_request` and `response` schemas.
- `create_request` is only created when create operation exists. All the rest is always generated.
- no data schema since we can now just use response.

This degrades a bit typing in the Python OpenAPI client (it struggles a bit to read what is required and what is not, but we don't care, it doesn't impact the tests).

Also, now we do not have required annotations on the response schema. But they were probably not needed anyway. Indeed, in such contexts we just return all the record, so in a way, either all or none of the fields should be marked as required. As we were re-using required fields of the create schemas, not all fields were marked, and this can be confusing. In the current state, we prefer no information over confusing information.

Still need to fix the docs.

> [!NOTE]
> Used a common schema for audit fields to gain some extra space 🙂

krilor:
> No impact anyway, it is only a tool during testing and no test is broken.

Hard disagree. Having an OpenAPI schema that lends itself to generate a nice client is a _feature_. We should not think of our use, but rather those that will use the system.

docs-dev/diagrams/openapi-schemas.drawio.png
krilor, line 1: This diagram seems off. Response schema should cover all properties?

docs-dev/openapi-schemas.md
krilor, line 1: Just noticed that this is in `docs-dev`. We should have this as part of `docs` and write it as public/user docs. How we structure our jsonschemas in the openapi doc is relevant information for thos integrating with the system

krilor (changes requested):

enzo-crance-statnett:
> Having an OpenAPI schema that lends itself to generate a nice client is a feature

As documentation is considered very high value in the rest of the project, I assumed good OpenAPI docs was the metric of success here. It is true I have not considered the fact people would integrate with our system by also using automated generators.

-----

But having docs rendered correctly yet a crappy client generated looks fishy. It suggests Elements can parse our docs correctly, but not the client generator. I tried [another bunch of generators](https://github.com/OpenAPITools/openapi-generator) outputting Python, Haskell and Kotlin, to get an idea of whether it is our OpenAPI docs that use (too) advanced referencing features, or if we just got unlucky and picked a bad generator a while ago.

Here are two examples, AP-ES response and SPG history response:

## APES response

What our generator outputs:

```python
AccountingPointEnergySupplierResponse:
    accounting_point_id: Union[Unset, int]
    energy_supplier_id: Union[Unset, int]
    valid_from: Union[Unset, str]
    valid_to: Union[None, Unset, str]
```

What the other generators output:

```python
AccountingPointEnergySupplierResponse(BaseModel):
    accounting_point_id: StrictInt
    energy_supplier_id: StrictInt
    valid_from: StrictStr
    valid_to: Optional[StrictStr]
```

```haskell
data AccountingPointEnergySupplierResponse = AccountingPointEnergySupplierResponse
  { accountingPointEnergySupplierResponseAccountingUnderscorepointUnderscoreid :: Int
  , accountingPointEnergySupplierResponseEnergyUnderscoresupplierUnderscoreid :: Int
  , accountingPointEnergySupplierResponseValidUnderscorefrom :: Text
  , accountingPointEnergySupplierResponseValidUnderscoreto :: Maybe Text
  }
```

```kotlin
data class AccountingPointEnergySupplierResponse (
    val accountingPointId: kotlin.Int,
    val energySupplierId: kotlin.Int,
    val validFrom: kotlin.String,
    val validTo: kotlin.String? = null
) {}
```

## SPG history response

What our generator produces:

```python
ServiceProvidingGroupHistoryResponse:
    service_providing_group_id: int
    name: Union[Unset, str]
    status: Union[Unset, ServiceProvidingGroupStatus]
    service_provider_id: Union[Unset, int]
    recorded_at: Union[Unset, str]
    recorded_by: Union[Unset, int]
    id: Union[Unset, int]
    replaced_by: Union[None, Unset, int]
    replaced_at: Union[None, Unset, str]
```

What the others produce:

```python
ServiceProvidingGroupHistoryResponse(BaseModel):
    name: Annotated[str, Field(strict=True, max_length=128)]
    status: ServiceProvidingGroupStatus
    service_provider_id: StrictInt
    id: StrictInt
    recorded_at: StrictStr
    recorded_by: StrictInt
    service_providing_group_id: StrictInt
    replaced_by: Optional[StrictInt]
    replaced_at: Optional[StrictStr]
```

```haskell
data ServiceProvidingGroupHistoryResponse = ServiceProvidingGroupHistoryResponse
  { serviceProvidingGroupHistoryResponseName :: Text
  , serviceProvidingGroupHistoryResponseStatus :: ServiceProvidingGroupStatus
  , serviceProvidingGroupHistoryResponseServiceUnderscoreproviderUnderscoreid :: Int
  , serviceProvidingGroupHistoryResponseId :: Int
  , serviceProvidingGroupHistoryResponseRecordedUnderscoreat :: Text
  , serviceProvidingGroupHistoryResponseRecordedUnderscoreby :: Int
  , serviceProvidingGroupHistoryResponseServiceUnderscoreprovidingUnderscoregroupUnderscoreid :: Int
  , serviceProvidingGroupHistoryResponseReplacedUnderscoreby :: Maybe Int
  , serviceProvidingGroupHistoryResponseReplacedUnderscoreat :: Maybe Text
  }
```

```kotlin
data class ServiceProvidingGroupHistoryResponse (
    val name: kotlin.String,
    val status: ServiceProvidingGroupStatus,
    val serviceProviderId: kotlin.Int,
    val id: kotlin.Int,
    val recordedAt: kotlin.String,
    val recordedBy: kotlin.Int,
    val serviceProvidingGroupId: kotlin.Int,
    val replacedBy: kotlin.Int? = null,
    val replacedAt: kotlin.String? = null
) {}
```

In both cases, we have these `Unset` almost everywhere, whereas the other generators understand correctly which values are always there and which are not. So fortunately it is not our spec which is bad, it is the generator.

----

All in all, I focused too much on the docs only, not caring much about the OpenAPI client out of ignorance, but in the end, having good docs proves to be a good metric, as most generators seem to follow in quality. It just looks like we have made an unlucky choice a while ago.

PR #236, opened 2025-06-26
feat: more fine-grained controllable unit lookup (FLEX-654)

enzo-crance-statnett:
This PR implements the changes to the Controllable Unit Lookup operation designed in https://github.com/elhub/flex-information-system/pull/233:

- changed the OpenAPI specification of the response;
- updated the SQL function backing the operation;
- introduced more input validation in the backend to distinguish more failure cases;
- updated tests to reflect the table of cases in https://github.com/elhub/flex-information-system/pull/233.

Draft because changes to docs and UI should be done as well before it is ready.


enzo-crance-statnett:
Updated UI

Screenshot 2025-06-27 125235

db/api/controllable_unit_lookup.sql
```diff
@@ -1,47 +1,126 @@
 --liquibase formatted sql
 -- Manually managed file
 
+-- changeset flex:api-controllable-unit-lookup-check-controllable-unit-exists endDelimiter:-- runAlways:true
+CREATE OR REPLACE FUNCTION
+api.controllable_unit_lookup_check_controllable_unit_exists(
```
krilor, line 6: Should we consider renaming this to remove the "controllable_unit_lookup_check" prefix? Like isn't this a more "general purpose" function?

db/api/controllable_unit_lookup.sql
```diff
@@ -1,47 +1,126 @@
...
+    SELECT EXISTS (
+        SELECT 1
+        FROM flex.controllable_unit AS cu
+        WHERE cu.business_id = l_controllable_unit_business_id::uuid
+    );
+$$;
+
+-- changeset flex:api-controllable-unit-lookup-check-end-user-matches-controllable-unit endDelimiter:-- runAlways:true
+CREATE OR REPLACE FUNCTION
+api.controllable_unit_lookup_check_end_user_matches_controllable_unit(
```
krilor, line 22: Having these "duplicate" functions for CU and AP is pointing at a possibility to simplify the logic a bit?
The "existance check" for CU should rather return the accounting point if successful?

So if doing CU lookup with CU, then the steps would be

1. currentControllableUnitAcoountingPoint -> accounting point buisness id | null
2. accountingPointEndUserMatches

db/api/controllable_unit_lookup.sql
```diff
@@ -1,47 +1,126 @@
...
     end_user_id bigint,
-    technical_resources jsonb
+    controllable_units jsonb
 )
 SECURITY DEFINER
 LANGUAGE plpgsql
 AS $$
 DECLARE
     l_event_data jsonb;
+    l_ap_id bigint;
```
krilor, line 89: One reflection I'm making while reading this is that this function should have been limited to just list controllable units, while getting the eu and ap info should have been part of the "pre checks".

enzo-crance-statnett:
Changes made following the comments related to the database structure and the way we get the data.

HTTP 500 gives less information, data validation returns AP and EU IDs on success, and CU lookup function only fetches CU as the rest is done before.

backend/data/data.go
```diff
@@ -268,51 +265,94 @@ func (data *api) controllableUnitLookupHandler(
...
+
+	endUserID, err := queries.ControllableUnitLookupCheckEndUserMatchesAccountingPoint(
+		ctx, endUserBusinessID, accountingPointBusinessID,
 	)
-	if errorCode != nil {
-		writeErrorToResponseWriter(w, *errorCode, *errorMsg)
+	if err != nil {
+		slog.InfoContext(
+			ctx, "end user does not match accounting point",
+			"end_user_business_id", endUserBusinessID,
```
krilor, line 318: Logging fødselsnummer (end_user_business_id) in clear text is not something we want to do.
I think we should just remove this log entry.

backend/data/data.go
```diff
@@ -646,3 +569,10 @@ func (data *api) notFoundHandler(w http.ResponseWriter, req *http.Request) {
 		Message: "Not Found " + req.URL.Path,
 	})
 }
+
+// writeInternalServerError writes a generic HTTP 500 response.
+func writeInternalServerError(w http.ResponseWriter) {
+	writeErrorToResponseWriter(w, http.StatusInternalServerError, errorMessage{ //nolint:exhaustruct
+		Message: "try again later",
```
krilor, line 576: try - Try?

krilor (changes requested):
Looking good. Remove one logline and we are good to go!

PR #244, opened 2025-07-03
feat(db): new "as-of" definition of technical resource (FLEX-660)

enzo-crance-statnett:
This PR redefines the technical resource view by using the as-of method, computing the as-of timestamp from CUSP for service providers.

Made the choice to ensure we have a row for the current party in the as-of view, so as to always be able to ask what the as-of timestamp is for a given CU. By default, current timestamp.

Still work in progress.

### DIFF
--- a/db/flex/changelog.yml
+++ b/db/flex/changelog.yml
@@ -108,9 +108,6 @@ databaseChangeLog:
       file: ./controllable_unit_service_provider.sql
       relativeToChangelogFile: true
   - include:
-      file: ./controllable_unit_as_of.sql
-      relativeToChangelogFile: true
-  - include:
       file: ./notification.sql
       relativeToChangelogFile: true
   - include:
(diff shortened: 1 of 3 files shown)


db/api/technical_resource.sql
```diff
@@ -35,24 +35,35 @@ WITH (
...
+            controllable_unit_id,
+            details,
+            name,
+            recorded_by,
+            record_time_range
+        FROM flex.technical_resource_history
+    ) AS trh
+        INNER JOIN flex.controllable_unit_as_of AS cu_asof
+            ON trh.controllable_unit_id = cu_asof.controllable_unit_id
+                AND cu_asof.party_id = flex.current_party()
```
krilor, line 65: I wonder if the "select pattern" used in RLS is also benefitial here? 🤔

From

```
AND cu_asof.party_id = flex.current_party()
```

to

```
AND cu_asof.party_id = (select flex.current_party())
```

PR #253, opened 2025-07-08
feat: add entity party (FLEX-641)

enzo-crance-statnett:
This PR introduces an _entity_ party type for future operations concerning an
organisation entity.

This means a user with role `flex_entity` can now either be a _bare entity_
(that is, one that has not assumed a party yet) or a user connected as the
entity of an organisation and who wants to act as the organisation itself.

JWT Bearer login was not splitting correctly the case where a party was given
and the bare entity case. It is also corrected in this PR.


krilor:
From what I remember from our discussion is that we landed on the "organization" party type after a bot of back-and-forth. Entity was the runner up but discarded since it is confusing to use the same name etc. What made you go with entity in the end?

enzo-crance-statnett:
I liked the idea of not having to create an extra role. That would require updates in quite a number of places and I'm not sure how to do the migration in a clean way. In addition, in some sense, _entity_ "typechecks" a bit more in my head than _organisation_, because you could theoretically create such a party for an entity that is a person and then we would need an extra type and duplicate some stuff.

Indeed, now the challenge is to document this very clearly, but I also remember that this solution was also an extra push towards having to do everything as a _party_ in the system, which is an idea I liked. We also said the user-facing text in the UI could be improved, which is not part of this PR (yet? I can think about adding that next).

The thing that could have been tricky is to tell both cases from each other (bare entity / entity assuming this new party), but we have a simple way to do it, by checking whether the current party exists.

krilor:
> I liked the idea of not having to create an extra role.

This is not really a valid concern for such a fundamental building block. We should do whatever changes are neccessary to make a "clean" system.

> could theoretically create such a party for an entity that is a person and then we would need an extra type and duplicate some stuff.

Yes. We discussed this and in my head we agreed that 1) orgs and persons are fundamentally a bit different (e.g. access policies) so it makes sense to split them 2) we probably never need the person stuff (???)

PR #254, opened 2025-07-09
docs: ready-for-market.md

kelvestad:


Brynhildur-Eddudottir-Statnett:
Agree with the new structure, added some "pirk".

PR #257, opened 2025-07-10
feat: scopes in party membership resource (FLEX-639)

enzo-crance-statnett:
This PR adds a `scopes` array field to the party membership resource to enable privilege delegation and entity administration later on. As a side effect, the resource has been made updatable so that scopes can be changed. For now, only FISO can do it.

Not adding a check constraint to ensure the content of scopes for now, as we do not really know what we will put there eventually. Adding scopes one by one through migrations sounds like a heavy process. I would prefer to figure out a sensible set of scopes and add the restriction in one go.

I am therefore starting with a soft, frontend-only restriction that allows putting either empty scopes (default) or `admin` as the only scope in the array. We could add something like `read_only` but we do not have a planned use for it soon. The feature is kind of minimal ATM.

To be refined in a later task that uses the scopes, for instance [FLEX-642](https://elhub.atlassian.net/browse/FLEX-642).

[FLEX-642]: https://elhub.atlassian.net/browse/FLEX-642?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


enzo-crance-statnett:
Input page

Screenshot 2025-07-10 130804

List

Screenshot 2025-07-10 130816

List in assume party page (we won't use an `admin` scope for the FISO party type, but it is just a way to illustrate the feature)

Screenshot 2025-07-10 130933

krilor (changes requested):
I do not agree with the idea of leaving scopes empty. Scopes should be treated as an allow-list, not a "modifier", IMHO. Empty = no scope = no access. Therefore, we must figure out what to call the scope we have had until now and add that as part of the migration.

In turn, null scope makes no sense since having a membership with no access is invalid.

The scope we have had until now is pretty powerful (lots of access) and should probably not be a default for new memberships tho.

krilor:
Isnt there also a bit of docs missing for this. Would expect this to be important to document in auth docs.

This PR also does not seem to suthorize based on/use the default scope. Shouldn't it?

enzo-crance-statnett:
> Shouldn't scopes be a finite list of possible values [...]?

I agree with the fact it should be a fixed enum eventually. My point was just to optimise for change in the beginning. Anyway, I have set a fixed list with `simple`, `admin`, `readonly`. `simple` replaces `resources` to signal that it is our "simple" access control that does not do anything advanced based on the scope.

> This PR also does not seem to authorize based on/use the default scope. Shouldn't it?

Indeed. Updated the RLS everywhere so that they just concern the `simple` scope. Except entity (that does not have scopes) and FISO (where scopes should not really matter). This means a party membership with no scope should now give almost zero access.

docs/technical/auth.md
```diff
@@ -378,6 +378,17 @@ grant_type=urn:ietf:params:oauth:grant-type:token-exchange
...
 access the API.
 
+### Party membership & scopes
+
+In our system, entities are made _members_ of parties to allow them to assume
+the parties and act on behalf of them.
+The membership relation comes with a list of _scopes_ (inspired by OAuth 2.0)
+that allow fine-tuning the access control when the entity assumes the party.
+Scopes drive what kind of resources the user is allowed to see or operations
+they can apply.
```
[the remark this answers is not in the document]
enzo-crance-statnett: Iterating on this:

| Scope         | Description                                                                |
|---------------|----------------------------------------------------------------------------|
| `data:read`   | Read-only access to business resources (CU, SPG, ...)                      |
| `data:manage` | Write access to business resources                                         |
| `auth:read`   | Read-only access to authentication resources (entity, party, clients, ...) |
| `auth:manage` | Write access to authentication resources                                   |

Maybe a role without `auth:read` could still see some auth resources, but only their own data and public data, not the rest. `admin` would then correspond to `auth:manage`.

enzo-crance-statnett:
Implemented more precise scopes splitting between data and auth, and read and manage. Stateful operations are asssociated to `manage`. I was unsure how to define `use` exactly, it probably doesn't fit our model.

Tests must still be updated and currently authentication is broken because entities cannot read parties (the scopes work with party memberships so entities are scope-less). I am trying to make a common model with entity client scopes so we can achieve both scope features with the same mechanism.

> [!NOTE]
> `auth:manage` + `flex_organisation` will correspond to the entity admin.

krilor (commented):
Looks like i had a couple pending inline comments that I didnt add the other day. One of them address the read vs use distinction.

.golangci.yml
```diff
@@ -13,9 +13,6 @@ linters:
     # nlreturn and wsl add lots of whitespace rules that are annoying
     - nlreturn
     - wsl
-    - wsl_v5
```
krilor: These are there in new versions of golangci. Upgrade and add them again?

db/flex/scopes.sql
```diff
@@ -0,0 +1,66 @@
...
+);
+
+-- changeset flex:validate-scope endDelimiter:; runOnChange:true
+ALTER DOMAIN flex.scope
+DROP CONSTRAINT IF EXISTS check_scope_resource;
+ALTER DOMAIN flex.scope
+ADD CONSTRAINT check_scope_resource CHECK (
+    split_part(value, ':', 3) IN (
+        '', -- no resource in scope is ok
+        'accounting_point',
```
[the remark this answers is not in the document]
enzo-crance-statnett: The other scopes exist but indeed we do not plan to use them.

db/flex/party_membership_migrations.sql
```diff
@@ -7,31 +7,16 @@
 ALTER TABLE flex.party_membership
 DISABLE TRIGGER USER;
 ALTER TABLE flex.party_membership
-ADD COLUMN scopes text [];
+ADD COLUMN scopes flex.scope [];
 
 UPDATE flex.party_membership
 SET scopes = '{manage:data, manage:auth}';
```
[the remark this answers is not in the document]
enzo-crance-statnett: The null check is added right after updating the table. I added the empty check though, this one was missing.

backend/auth/accesstoken.go
```diff
@@ -26,6 +28,8 @@ func (a accessToken) Validate() error {
 	val.Check(a.Role == "flex_entity" || a.PartyID > 0, "party_id is not set")
 	val.Check(!a.IsExpired(), "token has expired")
 	val.Check(a.ExternalID != "", "external_id is empty")
+	val.Check(a.Scope != nil, "scopes are not set")
```
krilor, line 31: NIT: scopes is now singular scope so error should be "scope is ..."

backend/auth/auth.go
```diff
@@ -120,17 +126,28 @@ func (auth *API) TokenDecodingMiddleware(
 
 		if authHeader == "" && err != nil {
 			// Empty auth header and missing cookie means anonymous user.
-			rd := &RequestDetails{role: "flex_anonymous", externalID: ""}
+			rd := &RequestDetails{
+				role:       "flex_anonymous",
+				externalID: "",
+				scope: scope.List{
```
krilor, line 132: NIT: Since we have defaultEntityScope we might also want defaultAnonymousScopes ?

backend/auth/auth.go
```diff
@@ -1274,15 +1415,21 @@ func (auth *API) jwtBearerHandler(
...
 
+	//nolint:nestif
 	if grant.Subject != nil {
 		// entity wants to assume a party
 		// we must first "log in" the entity to give it privileges in the database
-		rd := &RequestDetails{role: "flex_entity", externalID: externalID}
+		rd := &RequestDetails{
+			role:       "flex_entity",
+			externalID: externalID,
+			scope:      scope.List{},
```
krilor, line 1431: NIT: We might want to explain why we don't need to use defaultEntityScope here (?)

krilor (approved):

PR #258, opened 2025-07-14
feat: entity lookup (FLEX-640)

enzo-crance-statnett:
This PR introduces the entity lookup operation.

Tested ~~and supported in the UI~~.

### Input/Output

It expects as input the business ID of an entity that may or may not exist, along with a name and an entity type in case it does not exist and must be created.

The operation returns the technical ID of the entity. If the entity did not exist, it is created internally with the provided data. If the entity existed, then name and type are ignored.

Existing entity returns HTTP 200, newly created returns HTTP 201.

### Authorisation

The only authorised role for this operation is FISO as of now, because entity party / party membership scopes have not been merged yet. This will be easy to update as the role filter is in the backend for this operation, like for CU lookup.

### Notes

- For now, person entities can only be created with person number. We can change this to e-mail later, or allow both.

- It was not possible to use the `QUERY` HTTP method as we planned, because it would have not been possible to document it. Indeed, the current version of OpenAPI does not support this method. The feature got merged into the trunk for version 3.2 that is not released yet (cf https://github.com/OAI/OpenAPI-Specification/pull/4514/files). For uniformity, I went with `POST` and `/lookup` like for CU lookup. We can move the endpoints later.


krilor (approved):
Consider removing the lookup and result page. Would we ever need that? Wouldn't this functionality just be some kind of background thing used in the entity membership flow, rather than a separate page?

Also - shouldn't this be documented on the entity docs?

PR #259, opened 2025-07-17
feat: organisation party can change clients on owning entity (FLEX-642)

enzo-crance-statnett:
This PR introduces an RLS policy allowing users acting as an _organisation party with an `admin` scope_ to manage entity clients on the org entity owning their party.

Draft because dependent on [org party](https://github.com/elhub/flex-information-system/pull/253) and [party membership scopes](https://github.com/elhub/flex-information-system/pull/257) PRs. Maybe better to wait for merge of these before reviewing the current PR.


docs/resources/entity_client.md
```diff
@@ -74,7 +74,9 @@ No policies.
 
 #### Organisation
 
-No policies.
+| Policy key | Policy                                                                               | Status |
+|------------|--------------------------------------------------------------------------------------|--------|
+| ECL-ORG001 | Read, create, update and delete clients on the entity owning the organisation party. | DONE   |
```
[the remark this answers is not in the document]
enzo-crance-statnett: As discussed, we are adding a check for now to restrict the policy to users having a person entity. We will need additional checks to verify they have not logged in with an entity client, but we agreed to postpone that.

docs/resources/entity_client.md
```diff
@@ -74,7 +74,9 @@ No policies.
 
 #### Organisation
 
-No policies.
+| Policy key | Policy                                                                                                         | Status |
+|------------|----------------------------------------------------------------------------------------------------------------|--------|
+| ECL-ORG001 | Read, create, update and delete clients on the entity owning the organisation party, when the user is a human. | DONE   |
```
krilor, line 79: NIT: To be very picky, the "human" part is only really needed for CUD and not R.

docs/resources/entity_client.md
```diff
@@ -95,10 +97,10 @@ For party type abbreviations, check [the auth docs](../technical/auth.md#party-m
...
 |---------------|------|-----|----|----|------|----|----|----|----|-----|
 | id            |      | R   | R  | R  | R    | R  | R  | R  | R  | R   |
-| entity_id     |      | R   | R  | R  | R    | R  | R  | R  | R  | R   |
-| name          |      | R   | R  | R  | R    | R  | R  | R  | R  | R   |
-| client_id     |      | R   | R  | R  | R    | R  | R  | R  | R  | R   |
-| client_secret |      | R   | R  | R  | R    | R  | R  | R  | R  | R   |
-| public_key    |      | R   | R  | R  | R    | R  | R  | R  | R  | R   |
+| entity_id     |      | R   | R  | R  | R    | R  | R  | R  | R  | RC  |
+| name          |      | R   | R  | R  | R    | R  | R  | R  | R  | RCU |
+| client_id     |      | R   | R  | R  | R    | R  | R  | R  | R  | RU  |
```
krilor, line 102: client_id should not be updateable it is a read-only field - a generated id.

krilor (changes requested):
One thing that we must revisit is 

```
ECL-ENT001 | Read, create, update and delete their own clients.
```

Shouldn't the human bits also be part of that policy? Not saying we need to fix that as part of this PR, tho. Requesting changes to remove the U on entity_client.client_id

db/flex/entity_client_rls.sql
```diff
@@ -18,3 +18,33 @@ CREATE POLICY "ECL_FISO001" ON entity_client
...
+-- RLS: ECL-ORG001
+CREATE POLICY "ECL_ORG001" ON entity_client
+FOR SELECT
+TO flex_organisation
+USING (
+    entity_id = (SELECT flex.current_party_owner())
+);
+-- RLS: ECL-ORG002
+CREATE POLICY "ECL_ORG002" ON entity_client
+FOR ALL
```
krilor, line 45: We should probably not have FOR ALL here to avoid that this policy is active on selects?

krilor (changes requested):

krilor (approved):

PR #261, opened 2025-07-18
feat(ui): add CU lookup button on CUSP list on CU show page (FLEX-672)

enzo-crance-statnett:
This PR adds a CU lookup button on the CUSP list on the CU show page, so that SPs can get the end user technical ID a bit faster.

The button works the same as the one we already have in the CU list, except there the CU business ID gets autofilled on arrival, because we know we lookup by CU and not by AP.

Screenshot 2025-07-18 161154

Screenshot 2025-07-18 161247

### DIFF
--- a/frontend/src/controllable_unit/lookup/ControllableUnitLookupInput.tsx
+++ b/frontend/src/controllable_unit/lookup/ControllableUnitLookupInput.tsx
@@ -6,12 +6,13 @@ import {
   useNotify,
 } from "react-admin";
 import { Typography, Stack, Button, Box } from "@mui/material";
-import { useNavigate } from "react-router-dom";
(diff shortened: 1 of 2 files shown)


krilor (dismissed):
So Im guessing that the use case here is than an old sp wants to manage a CU where they previously had a contract? If so, shouldn't the button be about managing, and not about lookup (even tho what the SP technically need to do is to do lookup?)

Also the lookup page should have a few notes about what the page is used for ghat could help the user once they land there?

enzo-crance-statnett:
As they need the technical ID of the end user to be able to create the CUSP, and may no longer have it (since it is not an identifier related to business), they may need to re-run the lookup before, hence this button with autofill.

Updated the text on the lookup page to be more user-friendly.

PR #262, opened 2025-07-18
feat: reject new CUSP with same SP/EU as the current one (FLEX-676)

enzo-crance-statnett:


krilor (approved):
This seems to be a feature to protect the sp from themselves, in making equal consecutive contracts.
It is not a technical problem but rather a way to keep the data "clean"?

Should we give a hint to the user to update the existing contract? Also the validation rule description should make it clear that it is just an effort to keep things clean?

enzo-crance-statnett:
Indeed, while doing some QA, we identified that it was weird to have two CUSP with exactly the same settings following each other. It made sense to have this limitation even though it is not changing/fixing anything technically.

I changed the error message so it is explicit that the existing contract should be edited instead.

PR #268, opened 2025-08-25
docs: details on user creation flow in SP registration process (FLEX-637)

enzo-crance-statnett:


docs/processes/service-provider-registration-and-approval.md
```diff
@@ -46,3 +46,34 @@ be handled in the product application and qualification process.
...
+that allows at least to perform entity lookups and edit party memberships.
+
+In Step 19, Kari logs in as her own entity, then assumes the organisation party.
+She can then make the person numbers of her colleagues Ola and Tor members of
+the Service Provider party, so that they can also act on behalf of the company
+later on.
+Kari sets the privileges of her colleagues by adjusting which scopes they get on
+their party memberships.
+The operations include entity lookups to get technical entity IDs in the FIS
+from the person numbers, ensuring there is no double entity creation going on.
```
krilor, line 79: > ensuring there is no double entity creation going on

Ensuring no double entity is not the purpose of entity lookup. If we do not have any other technical constraints to ensure uniqueness on entity, then we should add that now.

docs/processes/service-provider-registration-and-approval.md
```diff
@@ -46,3 +46,34 @@ be handled in the product application and qualification process.
...
+  privileges to the user the FIS operator is in contact with in the context of
+  the current process (let us call her Kari Nordmann).
+
+In Step 16, the FISO creates a new person entity for Kari and makes it a member
+of the organisation party with administrator scopes, *i.e.*, a set of scopes
+that allows at least to perform entity lookups and edit party memberships.
+
+In Step 19, Kari logs in as her own entity, then assumes the organisation party.
+She can then make the person numbers of her colleagues Ola and Tor members of
+the Service Provider party, so that they can also act on behalf of the company
```
krilor, line 74: > on behalf of the company

Isnt that a bit impresise? "company" can easily be interpreted as "organisation", which is not what we are talking about here

docs/diagrams/service_provider_registration_details.drawio.png
krilor, line 1: I wonder if it would make it more clear if we introduce an actor/person in this diagram: Kari

?

Step 1. Kari logs in as herself and assumes the Solkraft Org Organisation party
Step 2. Kari does entity lookup to get technical id...
Step 3. Kari creates...

docs/processes/service-provider-registration-and-approval.md
enzo-crance-statnett: True, information is wrong here.

Double resources will be forbidden by the unique constraint on `entity.business_id`, but it is true entity lookup is rather a way to get the technical IDs in case the entities exist and they cannot read it, automating the creation behind when needed.

docs/diagrams/org_access_control_management.plantuml
krilor, line 1: Didn't we go a bit overboard on this diagram? I feel like we should slim it down for clarity. There is a lot of duplication of info and we can convey the concept more clearly by having a smaller diagram.

Suggestions:

1. Only show "Kari adds Ola". No need to show adding of multiple people.
2. Remove some of the fluff in the start. Starting point should not be a "Request access to FIS". All we need to know here is that the FISO wants to add Kari as admin and Kari wants to add Ola as a regular user. How that fits together with requesting access, signing contracts etc is scope for the process document(s) and not this document. I think the start should be step 4, possibly with a note on top.
3. The login and assume bits as regular users is covered by other parts of the docs and not something we need to duplicate here.
4. One possible improvement could be to add the action of giving identifier - Kari must give theirs to FISO and Ola must give theirs to Kari

docs/technical/auth.md
```diff
@@ -586,6 +586,87 @@ The following are a few example scopes.
...
+
+What we refer to as _administrator privileges_ here is a certain set of scopes
+on the party membership that makes the employee's person entity a member of the
+organisation party.
+The administrator is in charge for the addition of new users and the edition of
+party memberships and entity clients.
+The "administrator scopes" are therefore the ones sufficient to allow such
+operations, _i.e._, the following ones:
+
+* `use:data:entity:lookup`
```
krilor, line 638: Should we add the addtional "read:data" scope for completeness? I think it would be impossible with just these scopes.

docs/technical/auth.md
```diff
@@ -586,6 +586,87 @@ The following are a few example scopes.
...
+* `manage:data:party_memberships`
+* `manage:data:entity_client`
```
krilor, line 640: An improvement idea would be to have the list on a format like

```
* scope - reason
```

something like

```
* `use:data:entity:lookup` - to retrieve or create a technical identifier for the collegues they want to add in the system
```

docs/technical/auth.md
```diff
@@ -586,6 +586,87 @@ The following are a few example scopes.
 * `POST /api/v0/controllable_unit/lookup` requires `use:data:controllable_unit:lookup`.
   It is also covered by e.g. `manage:data` and `use:data:controllable_unit`.
 
+### User creation and access control management
```
krilor, line 589: This section is inserted at a bit weird place in the overall document, right in the middle of the 4 authz layers.

Suggested move:

bilde

We should also take it up one level to a h2/## heading

docs/diagrams/org_access_control_management.plantuml
enzo-crance-statnett: The idea was to show there were several parties, that the process did not only apply to SP for example, and that we could set read-only scopes.

But it is true there is duplication. I ended up simplifying the sequence diagram, while keeping all users mentioned in the other diagram, so that we keep information about read-only scopes etc, but the sequence is much lighter and easier to follow.

Login bits are only one step each now, to say who each user acts on behalf of, nothing more.

Added point nr.4 as well.

docs/technical/auth.md
enzo-crance-statnett: Agree. I wanted it to be after the place scopes are defined, so that it is not confusing, but it does not really only belong to "authorisation". Moved.

PR #269, opened 2025-08-27
feat(ui): support for party membership creation via entity lookup

enzo-crance-statnett:
This PR adds support in the UI to ease the process of adding an entity to a party when the user has to call entity lookup to get the entity's technical ID. It also grants the right to call entity lookup to organisation parties, as the feature was initially only tested on FISO.

The flow is that the user goes to the page of a party and gets its list of party memberships. On the top of the list, they can click a new button to create a membership _via lookup_.

Screenshot 2025-08-27 142333

This button saves the party ID in memory and brings the user to the entity lookup input page (roughly copied from the initial commit in https://github.com/elhub/flex-information-system/pull/258).

Screenshot 2025-08-27 142349

Once the lookup is done, the user is brought back to the party membership input page, with the party and entity already filled, so they just have to set the scopes and create the party membership.

Screenshot 2025-08-27 142417

There were many options to choose from when it comes to UI. Having a sub-form on the party membership page turned out to be too complicated technically and confusing. Here we have several pages, but at least the user does not have to "find" them. They just have to start from this additional button on the PM list component, and then gets guided through the pages by the submit buttons.

---

⚠️ One problem that showed up when working on this task is that organisation parties cannot read entities they create through lookup. This means when the entity is added to a visible party, then all components based on reading the entity do not work. For instance, the org party may see an empty name for an entity they added to one of the parties. Ideally we should fix this before merging. I suggest adding the following policy on entities for organisation parties:
```
ENT-ORG001 | Read all entities belonging to parties owned by the organisation.
```
This means the technical ID would be used during the party membership creation process, but once it is created, the organisation can see the entity's details.

---

Marking as a draft since dependent on https://github.com/elhub/flex-information-system/pull/260 to function correctly and be tested.


backend/data/data.go
```diff
@@ -473,7 +473,12 @@ func (data *api) entityLookupHandler(
 		return
 	}
 
-	if rd.Role() != "flex_flexibility_information_system_operator" {
+	allowedRoles := []string{
+		"flex_flexibility_information_system_operator",
```
krilor, line 477: This functionality is what I've coined "Party type check" in https://github.com/elhub/flex-information-system/blob/main/docs/technical/auth.md#overview

Should we consider making this a `auth` middleware at some point? It at least makes sense to me that we can "sprinkle in" party type checks where relevant.

krilor (approved):
Nice!

PR #270, opened 2025-08-28
feat: set default scopes when doing login through party ownership (FLEX-692)

enzo-crance-statnett:
This PR adds default scopes when the entity assumes a party they _own_.

We give the required scopes for the API to function correctly, among which read-write access to data (it makes sense to have some sort of regular access at least, when you own the party).

### DIFF
--- a/db/auth/functions.sql
+++ b/db/auth/functions.sql
@@ -82,9 +82,11 @@ begin
   select flex.current_entity() into _entity_id;
 
   if exists (
+    -- entity is member of party
     select 1 from flex.party_membership pm
     where pm.entity_id = _entity_id and pm.party_id = _party_id
   ) or exists (
+    -- entity owns party
     select 1 from flex.party as p
     where p.id = _party_id and p.entity_id = _entity_id
   ) then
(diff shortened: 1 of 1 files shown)


db/auth/functions.sql
```diff
@@ -95,9 +97,16 @@ begin
...
   if eid is not null then
     select p.role into role from flex.party p where p.id = _party_id;
+
+    -- get scopes from party membership
     select pm.scopes into l_scopes
     from flex.party_membership pm
     where pm.entity_id = _entity_id and pm.party_id = _party_id;
+
+    -- if not existing, default scopes for party ownership
+    if l_scopes is null or array_length(l_scopes, 1) = 0 then
```
krilor, line 107: NIT: Reading this makes me stop and pause. Thinking "so, if scopes are empty we just give it all-the-access?". This ofcourse _relies_ on the fact that we have a constraint that party_membership.scopes is not null and not empty. This is probably never going to change, but it made me think of this article:

https://surfingcomplexity.blog/2024/03/26/the-problem-with-invariants-is-that-they-change-over-time/

> Implicit invariants are, by definition, impossible to enforce explicitly. They are time bombs. And they are everywhere.

How to fix? Branch earlier - If membership, do this. If ownership do this

krilor (dismissed):

krilor (approved):

PR #272, opened 2025-08-29
feat: add party_id and scopes fields to entity client (FLEX-694)

enzo-crance-statnett:
This PR adds `party_id` and `scopes` fields to the entity client resource, without using them for anything yet. This allows merging the boilerplate before we focus on the auth bits, spotting some rules we can already come up with along the way.


enzo-crance-statnett:
For instance, we should restrict the `party_id` field from the beginning, to contain only parties that the entity can actually assume. Otherwise it creates an "invalid" client, that should not really be dangerous (thanks to the verifications we make on party assume) but is guaranteed to never work, which is not really what we want to have in the system.

PR #273, opened 2025-09-01
feat: assign entity scopes from entity client (FLEX-695)

enzo-crance-statnett:
This PR adds logic to set scopes for an entity from the entity client they chose to log in, if it applies. Otherwise they get standard RW authorisation as a default.


enzo-crance-statnett:
Marking as ready for review (the scopes can stay untouched for now where the todo comment is), and opening a new PR for the proposed identity changes.

PR #274, opened 2025-09-02
feat: add entity client ID to identity

enzo-crance-statnett:
This PR proposes to add an optional entity client ID to the identity table.

We should also evaluate whether showing the entity client ID in userinfo and in the API is relevant. For now, this is only an internal change.


krilor (dismissed):
🔥Nice!

PR #275, opened 2025-09-03
feat: backend scope intersection functionality (FLEX-696)

krilor:


krilor:
> I guess the duplicates won't hurt

What duplicates? There is functionality to remove actual duplicates but it does not handle overlapping scopes or anything like that.

> do the intersection in the backend

Yep. It is the auth module that is responsible for establishing the access for a session, so it makes sense to me. Also, doing this type of thing in sql is painful IMHO.

PR #303, opened 2025-09-25
feat(db): implement history policy changes for party membership (FLEX-728)

enzo-crance-statnett:
cf https://github.com/elhub/flex-information-system/pull/302

### DIFF
--- a/test/api_client_tests/test_party_membership.py
+++ b/test/api_client_tests/test_party_membership.py
@@ -231,6 +231,17 @@ def test_ptym_org(sts):
     # RLS: PTYM-ORG002
     # they can read history on such PM
 
+    # TODO: edit when the manage:data:party_membership_history scope exists
+    u = update_party_membership.sync(
+        client=client_fiso,
+        id=cast(int, org_pm.id),
+        body=PartyMembershipUpdateRequest(
+            scopes=[AuthScope.MANAGEDATA],
+        ),
+    )
+    assert not (isinstance(u, ErrorMessage))
+    client_org = sts.get_client(TestEntity.TEST, "ORG", reset=True)
+
     pmhs = list_party_membership_history.sync(
         client=client_org,
         party_membership_id=f"eq.{cast(int, pms[0].id)}",


backend/auth/middleware.go
```diff
@@ -82,6 +82,8 @@ func CheckScopeForRequest(module string, next http.Handler) http.Handler {
 		requestPath := req.URL.Path
 		requestPath = strings.TrimPrefix(requestPath, "/")
 		requestPath = strings.TrimSuffix(requestPath, "/") // Remove trailing slash if present
+		// if requesting a history resource, base the scope on the main resource
+		requestPath = strings.TrimSuffix(requestPath, "_history")
```
enzo-crance-statnett, line 86: `manage:data:party_membership` now gives `read:data:party_membership_history`. I think in pretty much all cases it makes sense that opening a resource also opens access to the history. History is baked in almost all resources and I don't see a case where we would want to make history access forbidden.

BUT it is a special case we add there, it should probably be documented etc. and we may not like it because it is a source of "surprise".

An alternative is to add this specific history scope everywhere (UI, test data, etc). I think it is heavy to have to add history explicitly whenever we give access to a specific resource. But it is more predictable / less surprising.

I chose the current approach because it was the cheapest one in terms of changes to make. I can either document the current change better or switch to adding the specific scope if we end up preferring that other option.

backend/event/models/models.custom.go
enzo-crance-statnett, line 1: Changes in the backend models come from a previous PR, I just forgot to re-generate the code. Sorry for that.

backend/auth/middleware.go
krilor: This is an interesting suggestion but cannot be integrated in its current state. Missing docs (md and openapi) + there is not actual reason for doing it except "it makes sense in _most_ cases" . "Most cases" is not a good reason for such a default. We are loosing the ability to do granular access on history.

krilor (changes requested):

enzo-crance-statnett:
We can bypass the problem by giving a more general scope in the test for now.

This means scopes as we specified them in the API (`auth_scope` OpenAPI schema) do not give enough granularity to control read access on this history table. So organisation admins as we create them in the test data and in the UI cannot read PTYM history right now. We should definitely revisit the question later.

krilor (approved):

PR #305, opened 2025-09-25
feat: update SPPA and SPPS comment visibility and fix history policies (FLEX-728 / FLEX-743)

enzo-crance-statnett:
cf https://github.com/elhub/flex-information-system/pull/302

> [!WARNING]
> I discovered the base policies (not related to history) were not fully implemented for the SPPS comment resource.
>
> Basically, we were not filtering on the identity of the user, only on the visibility of the comment. As an SO, we wrote in the SPPSC-SO002 policy that you could see a comment when you are PSO on the suspension and the visibility allows it. We were actually only checking the visibility. It means that as SPPS has these extra restrictions on being a PSO etc, as an SO, you can currently possibly see the comments of a SPPS if they are open to your party or any party, without being able to see the suspension itself. Same for a SP not being the target of a suspension. This has to be corrected.
>
> However, if we filter correctly by PSO / suspended SP, then basically "same_party_type" comments do not make sense, because only 3 persons can read the comments of a suspension: the suspended SP, the suspending PSO, and possibly a FISO. Other SPs/SOs cannot read any comment. And I don't know whether this is functionally correct.
>
> I did not update the tests when I noticed this. Leaving the PR in draft until I get clarifications.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -4,6 +4,13 @@ This changelog tracks all changes that are visible in our API.
 
 <!-- markdownlint-disable MD013 -->
 
+## 26.09.2025
+
+* **Updated the `visibility` field's specification on the _SP product application comment_ and the _SP product suspension comment_ resources.**  
+  Comments now have only two possible visibilities: they can either be private
+  comments internal to a given party, or comments opened to any party related to
+  the application or suspension.
+
 ## 16.09.2025
 
 * **Renamed all `last_X` timestamp fields into `X_at`.**


enzo-crance-statnett:
After discussion, reducing visibility on comments to only two values, `same_party` and `any_involved_party`, documented in OpenAPI and tested.

This PR now concerns two tasks at once.

enzo-crance-statnett:
It turns out the history RLS required to check _both_ history and the main resource. It will have to be changed on other resources whose checks are based on history as well (SPPS and TR), but this can be done in later PRs: for SPPS it will be a bug fix, for TR it is planned that I change it as part of FLEX-728 and I will remember this when I do the change.

krilor (dismissed):

PR #308, opened 2025-09-26
fix(db): update RLS for SPPS when the check is based on history (FLEX-728)

enzo-crance-statnett:
This PR fixes the bug on history-based RLS described in https://github.com/elhub/flex-information-system/pull/305#issuecomment-3338580016 for SPPS and removes overlapping policies.

The existing test was in fact already the right one, but it was not testing the right policy.

Indeed, as an SO, if a suspension is yours, then you are suspending someone you have qualified before, as ensured by the validation rule (SPPS-VAL001). This means the rule giving history access on your own suspensions (the former SPPS-SO002) is _always_ covered by the rule giving history access on suspensions targeted to SPs you have qualified once (the former SPPS-SO004). So in the case where it is your own suspension, the first rule was short-cutting the second, which means the second was not really being tested.

By removing the shortcut, we simplify the RLS and the tests do not even have to change.

### DIFF
--- a/db/flex/service_provider_product_suspension_rls.sql
+++ b/db/flex/service_provider_product_suspension_rls.sql
@@ -68,7 +68,16 @@ TO flex_system_operator
 USING (procuring_system_operator_id = (SELECT flex.current_party()));
 
 -- RLS: SPPS-SO002
+GRANT SELECT ON service_provider_product_suspension_history
+TO flex_system_operator;
 CREATE POLICY "SPPS_SO002"
+ON service_provider_product_suspension_history
+FOR SELECT
+TO flex_system_operator
+USING (procuring_system_operator_id = (SELECT flex.current_party()));
+
+-- RLS: SPPS-SO003
+CREATE POLICY "SPPS_SO003"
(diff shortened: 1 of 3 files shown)


krilor (commented):
It is more-than-likely that I am missunderstand something here, but removing this policy has three downsides imho.

> SPPS-SO002 | Read history on their own SPPS. 

1. It makes it hard to reason about the policies. I have to jump through hoops to understand the the SP can read history on their own SPPS. This is not good. The existing policy is simple and the intention is very clear.
2. We are leaning on a validation rule that might change at some point. Say that validation rule changes in some way that makes our assumption here false - it would be close to impossible to remember or figure out that we would have to re-introduce this policy.
3. Performance. AFAIK the postgres planner will evaluate the policies in sorted order from least to most expensive. The removed policy is very cheap; evaluating the data vs a tiny STABLE function just accessing a current setting. The "shortcut" is in fact a very benefitial shortcut from a performance perspective.

For now, just going with "Comment" since I'm ready to be convinced otherwise.

krilor (approved):

PR #312, opened 2025-09-29
fix: make event policies on deletable resources history-based (FLEX-728)

enzo-crance-statnett:
This PR fixes RLS for the event resource so that we can still read create events and delete events even after removal of a row in a deletable resource.

### DIFF
--- a/db/flex/event_rls.sql
+++ b/db/flex/event_rls.sql
@@ -106,8 +106,8 @@ USING (
     )
 );
 
--- RLS: EVENT-SP002
-CREATE POLICY "EVENT_SP002" ON event
+-- RLS: EVENT-SP003
+CREATE POLICY "EVENT_SP003" ON event
 FOR SELECT
 TO flex_service_provider
 USING (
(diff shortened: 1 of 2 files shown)


db/flex/event_rls.sql
```diff
@@ -119,21 +119,21 @@ USING (
...
+-- RLS: EVENT-SP002
+CREATE POLICY "EVENT_SP002" ON event
 FOR SELECT
 TO flex_service_provider
 USING (
     source_resource = 'technical_resource'
     AND EXISTS (
         SELECT 1
-        FROM flex.technical_resource AS tr
+        FROM flex.technical_resource_history AS trh
```
krilor, line 130: flex.technical_resource_history has no rows when the TR has only been created (which is different behaviour from api.technical_resource_history btw).

I'm not able to see how this new policy would cover create events. I see that ther are asserts for them in the test cases so I guess I'm just missing something, but would love a short explanation of how that works.

I'm on thin ice here but I kindof assumed we had to do a "TR UNION ALL TRH".

krilor (dismissed):

db/flex/event_rls.sql
enzo-crance-statnett: It is exactly the reason behind one of the bugs I fixed in another recent PR. This definitely has to be fixed as well.

I guess we still see the create event because we update or delete the TR.

krilor (approved):
Nice!

PR #314, opened 2025-09-30
feat: load MGA data using Liquibase (FLEX-749)

enzo-crance-statnett:
This PR loads MGA data using Liquibase.

The test MGA required an SO as an owner, and this party required an entity, hence the extra line in the 3 files.
(I could have reused the existing entity though, IDK)


db/flex/changelog.yml
```diff
@@ -291,6 +291,24 @@ databaseChangeLog:
...
+  - changeSet:
+      id: mga-structure-data
+      author: flex
+      changes:
+        - loadUpdateData:
+            file: ./structure_data/mga.csv
+            relativeToChangelogFile: true
+            schemaName: flex
+            tableName: metering_grid_area
+            separator: ";"  # valid time ranges contain commas
```
krilor, line 303: We should have the same separator on all files.
We just need to quote the field?

https://docs.liquibase.com/reference-guide/change-types/loaddata

> quotchar
>
> The quote character for string fields containing the separator character.

krilor (approved):

PR #316, opened 2025-10-01
feat: add support for EIC-Y business ID type and use it for MGAs (FLEX-749)

enzo-crance-statnett:
This PR adds EIC-Y codes for grid areas. We were using EIC-X also for MGAs before, which is wrong.

The migration replaces the `X` with `Y` in the existing MGAs, but it has to take care of recomputing check digits and possibly changing the prefix so that it is correct.

For instance, changing `X` into `Y` in `42X-U----------D` (example of MGA business ID in the formerly used EIC-X format) turns the check digit into `-`, which is invalid. Checkable [here](https://www.entsoe.eu/data/energy-identification-codes-eic/code-generator/eic_key_generator). In such cases we need a procedure to update the prefix in a way that:
- is simple enough for us not to lose too much time on the task;
- potentially catches several rounds of `-` check digit;
- terminates, because we don't want an infinite loop in a migration.

I came up with the following solution: replace some characters with `X` (chosen randomly in the acceptable alphabet) until the check digit is valid.

TBH I haven't thought about the theoretical probability of the check digit being wrong several times, but I felt doing only one replace was not enough if you are very unlucky. And once I had the loop in place, why not loop on like ~10 characters. We early exit if everything is correct anyway. And now it's probably impossible to have any invalid code after that.

Draft because tests are broken in `main` (they still depend on hard-coded entity/party IDs in some places and we have been touching this data with our liquibase structure data updates) so I cannot ensure it is ready, even though I tested the migration manually and it looks right.

### DIFF
--- a/db/flex/business_id_type.sql
+++ b/db/flex/business_id_type.sql
@@ -10,7 +10,7 @@ CREATE TABLE IF NOT EXISTS flex.business_id_type (
     )
 );
 
--- changeset flex:validate-business-id endDelimiter:-- runOnChange: true
+-- changeset flex:validate-business-id endDelimiter:-- runOnChange:true
 CREATE OR REPLACE FUNCTION flex.validate_business_id(
     business_id text, business_id_type text
 )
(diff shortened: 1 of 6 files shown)


krilor:
> I came up with the following solution: replace some characters with `X` (chosen randomly in the acceptable alphabet) until the check digit is valid.

we could follow the same strategy as in the test data package to not reinvent the wheel and keep things the same. https://github.com/elhub/flex-information-system/blob/5f4c9cc893da8ab8ea3ea4b62d00ad68b65c9339/db/test_data/test_data.sql#L584C26-L584C41

Doesn't really matter tho

krilor (approved):

PR #321, opened 2025-10-06
feat: add SPG grid suspension resource (FLEX-760)

enzo-crance-statnett:
This PR will add the SPGGS resource.

### DIFF
--- a/docs/resources/service_providing_group_grid_suspension.md
+++ b/docs/resources/service_providing_group_grid_suspension.md
@@ -1,0 +1,108 @@
+# Service Providing Group Grid Suspension
+
+The relation allowing an impacted system operator to temporarily suspend a
+service providing group from delivering products.
+
+## Relevant links
+
+* [API Documentation](../api/v0/index.html#/operations/list_service_providing_group_grid_suspension)
+* [Download docx](../download/service_providing_group_grid_suspension.docx)
+
+## Fields
+
+| Name                                                                                                                  | Description                                                                       | Format                                                                                   | Reference                                                         |
+|-----------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------|------------------------------------------------------------------------------------------|-------------------------------------------------------------------|
+| <a name="field-id" href="#field-id">id</a>                                                                            | Unique surrogate identifier.                                                      | bigint<br/>Read only                                                                     |                                                                   |
+| <a name="field-impacted_system_operator_id" href="#field-impacted_system_operator_id">impacted_system_operator_id</a> | Reference to the impacted system operator suspending the service providing group. | bigint<br/>Non-updatable<br/><br/>Defaults to the system operator creating the resource. | [party.id](party.md#field-id)                                     |
+| <a name="field-service_providing_group_id" href="#field-service_providing_group_id">service_providing_group_id</a>    | Reference to the service providing group being suspended.                         | bigint<br/>Required<br/>Non-updatable                                                    | [service_providing_group.id](service_providing_group.md#field-id) |
+| <a name="field-reason" href="#field-reason">reason</a>                                                                | The reason for the suspension.                                                    | text<br/>One of: `breach_of_conditions`, `significant_alteration`, `other`<br/>Required  |                                                                   |
+| <a name="field-recorded_at" href="#field-recorded_at">recorded_at</a>                                                 | When the resource was recorded (created or updated) in the system.                | timestamp with time zone<br/>Read only                                                   |                                                                   |
+| <a name="field-recorded_by" href="#field-recorded_by">recorded_by</a>                                                 | The identity that recorded the resource.                                          | bigint<br/>Read only                                                                     |                                                                   |
+
+## Validation Rules
+
+No validation rules.
+
+## Notifications
+
+| Action                 | Recipient | Comment                                                             |
+|------------------------|-----------|---------------------------------------------------------------------|
+| create, update, delete | SP        | Suspended SP                                                        |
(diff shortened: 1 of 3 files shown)


docs/resources/service_providing_group_grid_suspension.md
```diff
@@ -0,0 +1,108 @@
...
+
+No policies.
+
+#### System Operator
+
+| Policy key  | Policy                                                                                                    | Status |
+|-------------|-----------------------------------------------------------------------------------------------------------|--------|
+| SPGGS-SO001 | Create, read, update and delete their own SPGGS.                                                          | TODO   |
+| SPGGS-SO002 | Read history on their own SPGGS.                                                                          | TODO   |
+| SPGGS-SO003 | Read SPGGS targeted at any SPG for which they have qualified or verified a product application.           | TODO   |
```
maribjorn, line 81: Prequalified instead of "qualified"

db/flex/service_providing_group_grid_suspension.sql
```diff
@@ -0,0 +1,74 @@
...
+    reason text NOT NULL,
+    record_time_range tstzrange NOT NULL DEFAULT tstzrange(
+        localtimestamp, null, '[)'
+    ),
+    recorded_by bigint NOT NULL DEFAULT current_identity(),
+
+    CONSTRAINT service_providing_group_grid_suspension_reason_check CHECK (
+        reason IN (
+            'breach_of_conditions',
+            'significant_alteration',
```
krilor, line 21: There was a name-change in https://github.com/elhub/flex-information-system/pull/323#discussion_r2409685771 that must be taken into account here.

db/flex/service_providing_group_grid_suspension.sql
```diff
@@ -0,0 +1,74 @@
...
+LANGUAGE plpgsql
+AS
+$$
+BEGIN
+    IF NOT EXISTS (
+        SELECT 1
+        FROM flex.service_providing_group_grid_prequalification AS spggp
+        WHERE spggp.service_providing_group_id = NEW.service_providing_group_id
+        AND spggp.impacted_system_operator_id = NEW.impacted_system_operator_id
+        AND spggp.status = 'approved'
```
krilor, line 49: I think this condition is a bit too strict. The "business rule" is more like "is the SPG considered ready for market by the ISO".

Refs:
* https://elhub.github.io/flex-information-system/concepts/ready-for-market/#service-providing-group-checklist
* https://elhub.github.io/flex-information-system/resources/service_providing_group_grid_prequalification/#field-status

So the condition should me more like 

```sql
AND ( spggp.status in ('approved', 'conditionally_approved') or prequalified_at is not null)
```

db/flex/service_providing_group_grid_suspension_rls.sql
```diff
@@ -0,0 +1,181 @@
...
+TO flex_system_operator
+USING (
+    EXISTS (
+        SELECT 1
+        FROM flex.service_providing_group_product_application AS spgpa
+        WHERE spgpa.service_providing_group_id
+            = service_providing_group_grid_suspension.service_providing_group_id -- noqa
+            AND spgpa.procuring_system_operator_id
+            = (SELECT flex.current_party())
+            AND spgpa.status IN ('prequalified', 'verified')
```
krilor, line 106: Ref my other comment about "ready for market" should we not check the timestamps (prequalified_at, verified_at) here as well?

db/test_data/test_data.sql
```diff
@@ -812,7 +812,7 @@ BEGIN
     PERFORM test_data.add_accounting_points(
       accounting_point_prefix,
       so_mga_id,
-      null,
+      (eu_id, eu_id),
```
krilor, line 815: ?

docs/resources/service_providing_group_grid_suspension.md
```diff
@@ -0,0 +1,110 @@
+# Service Providing Group Grid Suspension
+
+The relation allowing an impacted system operator to temporarily suspend a
+service providing group from delivering products.
```
krilor, line 4: NIT: could use "delivering services.". Just a feeling that its a better word to use in this context.

docs/resources/service_providing_group_grid_suspension.md
```diff
@@ -0,0 +1,110 @@
...
+
+No policies.
+
+#### System Operator
+
+| Policy key  | Policy                                                                                                       | Status |
+|-------------|--------------------------------------------------------------------------------------------------------------|--------|
+| SPGGS-SO001 | Create, read, update and delete their own SPGGS.                                                             | DONE   |
+| SPGGS-SO002 | Read history on their own SPGGS.                                                                             | DONE   |
+| SPGGS-SO003 | Read SPGGS targeted at any SPG for which they have prequalified or verified a product application.           | DONE   |
```
krilor, line 83: Is this rule a bit strict? Isn't it relevant to see suspensions _when prequalifiying_?

I would expect the suspension to be similar to its sister-resource grid prequalification
https://elhub.github.io/flex-information-system/resources/service_providing_group_grid_prequalification/

krilor (changes requested):

db/test_data/test_data.sql
enzo-crance-statnett: This is related to testing. I needed accounting points with 2 different end users behind them.

I used the common EU as end user for APs from 0 to 1000. I could have used the new "Emil Post" user as well, but this requires some changes in the security token service and justfile to set up an entity client for client credentials. I went with the easiest change, but we are not tied to it.

PR #335, opened 2025-10-15
fix: updated assume party ux (FLEX-779)

sigridOC:
Moved the order so that the assume party is the rightmost element.

Added a singular person icon and green text with gray boarder on hover so it stands out a little ( see hover effect at the first button in the picture)

changed Chips design to separate them from buttons

image

### DIFF
--- a/frontend/src/AssumePartyPage.tsx
+++ b/frontend/src/AssumePartyPage.tsx
@@ -196,7 +196,7 @@ export const AssumePartyPage = () => {
             }
             disableSyncWithLocation
           >
-            <Datagrid bulkActionButtons={false}>
+            <Datagrid bulkActionButtons={false} rowClick={false}>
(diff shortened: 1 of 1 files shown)


krilor (dismissed):
These changes look good!

I have a few suggestions that we should _consider_ as part of the current task FLEX-779. Could be this PR or in a follow-up.

1. Set `link={false}` on the type column. It is a referencefield (and thus link) to the party which is a bit confusing. Having the link on the party name is enough and ok.
2. The behaviour of the "Parties you own" thing is slightly different from "Parties you belong to". The whole row is clickable. The behaviour should be similar. Can we figure out why and fix it?

enzo-crance-statnett (dismissed):

maribjorn (approved):
Looks very good!

PR #340, opened 2025-10-16
fix: upgrading frontend packages (FLEX-780)

sigridOC:
React hook rules ar violated. It seems that the React functions or hooks that are created within a rendering methods needs to be moved outside.

### DIFF
--- a/docs/processes/index.md
+++ b/docs/processes/index.md
@@ -73,7 +73,7 @@ also use the menu to find process details.
 ### Access control
 
 Entitled parties have permission to access data in the flexbility information
-system as described in the Authorization section for each [resource](../resources/).
+system as described in the Authorization section for each [resource](../resources/index.md).
 
 ### Controllable unit
 
--- a/docs/processes/service-providing-group-grid-suspension.md
+++ b/docs/processes/service-providing-group-grid-suspension.md
@@ -1,4 +1,4 @@
-# SPG grid suspension
+# SPG Grid Suspension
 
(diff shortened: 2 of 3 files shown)


enzo-crance-statnett (dismissed):

enzo-crance-statnett:
Not sure I follow the message.

If the linters / Megalinter are complaining about something and the build fails with the new versions of the packages, we should address it in the current PR.

Otherwise it is fine to merge and fix some IDE warnings later.

krilor (changes requested):
We should keep main "clean" and this seems to break CI, so this cannot me merged in current state. Fixing the errors must be part of the PR or fixed in another PR prior to merge here.

frontend/src/entity/client/EntityClientList.tsx
```diff
@@ -16,33 +16,43 @@ import { DateField } from "../../components/datetime";
...
+  />
+);
+
+const ListActions = (permissions: string[], id: any) => (
+  <TopToolbar>
+    {permissions.includes("entity_client.create") && <CreateButton id={id} />}
+  </TopToolbar>
+);
+
+const ActionListWrapper = ({
```
krilor: Why do we need a wrapper here? Any way we can avoid it? Looks like we don't need it other places.

sigridOC: Nice catch. Yes I think it is a bit redundant. Missed it during cleanup!

krilor (approved):

PR #341, opened 2025-10-16
feat: increase automation in comment history/view generation (FLEX-792)

enzo-crance-statnett:
This PR adds even more automation for comment resources, since they all look alike. The idea is just to have to add `comments: true` to a resource to get the whole set of comment files generated.

Draft for communication. We still miss OpenAPI/Markdown generation.

### DIFF
--- a/openapi/resources-future.yml
+++ b/openapi/resources-future.yml
@@ -1,0 +1,21 @@
+# used to generate resources.png
+# Add future resources here
+
+resources: []
+  # - id: future_resource
+  #   future: true
+  #   acronym: FR
+  #   name: Future Resource
+  #   summary: Yep, its the future
+  #   description: Some meaningful description of the resource
+  #   properties:
+  #     connecting_system_operator_id:
+  #       description: >-
+  #         Reference to the `party` that is the connecting system operator.
+  #         Derived from the `accounting_point_id`.
+  #       format: bigint
+  #       type: integer
+  #       readOnly: true
+  #       x-foreign-key:
+  #         resource: party
+  #         field: id
--- a/justfile
+++ b/justfile
@@ -368,7 +368,8 @@ template-to-openapi:
 resources-to-diagram:
     #!/usr/bin/env bash
     set -euo pipefail
-    cat openapi/resources.yml | .venv/bin/python3 local/scripts/resources_to_diagram.py | tee docs/diagrams/resources.plantuml
+    yq eval-all '. as $item ireduce ({}; . *d $item )' openapi/resources.yml openapi/resources-future.yml | \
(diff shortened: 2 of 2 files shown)


local/scripts/openapi_to_db.py
```diff
@@ -80,6 +82,79 @@ def fake_history_table_create_statement(resource, properties):
 """
 
 
+def base_comment_resource_data(base_resource):
```
enzo-crance-statnett: This dictionary can be turned into another template if it happens we must also define it in another script, so we don't repeat ourselves. I will see.

local/scripts/openapi_to_db.py
```diff
@@ -125,3 +200,39 @@ def fake_history_table_create_statement(resource, properties):
...
+
+            # generate files for the comment resource
+            if resource.get("comments", False):
+                comment_resource_file = f"{DB_DIR}/flex/{resource['id']}_comment.sql"
+                if not os.path.exists(comment_resource_file):
+                    j2.template(
+                        resource,
+                        "comment_resource.j2.sql",
+                        comment_resource_file,
+                    )
```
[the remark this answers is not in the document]
krilor: I _think_ i dissagree. One of the main benefits of generation is that to be able to maintain consistency _over time_. Fix once, then it propagates to all the resources. Should we add all this complexity just to get a slightly easier workflow than copypaste+searchandreplace? If that is what we are after, then maybe code generation is the wrong path here? We are not talking about _that_ many comment resources.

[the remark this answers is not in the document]
enzo-crance-statnett: Fixing stuff goes through migrations anyway, so you are forced to fix the existing resources "manually". But you can have consistency for _later_ resources by fixing the template for the next times you use it. Your favourite project manager has this `init` or `new` command that creates a project, module, whatever. With such an approach, generation is used more as a kind of starting blocks. I was trying to think along these lines, I am not sure if this is best either. 🤔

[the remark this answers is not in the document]
enzo-crance-statnett: The "new" thing I tried here is to also generate the main table, not only history and views. And the issue we have with repeated generation of tables is the migrations along the way. The generated file will always differ because of Liquibase, for example the line with the checksum.

Maybe it is best to keep generation to history and views. I will explore this a bit more, and go back to manual if we do not come to a good conclusion.

local/scripts/openapi_to_db.py
enzo-crance-statnett: Done ✅

openapi/resources-future.yml
krilor: Is deleting this file intentional?

krilor: Ah. Well its not dead code but a feature we have used very little.
Id prefer if we avoid removing "unrelated" features as part of this PR. Can we?

PR #342, opened 2025-10-17
chore: upgrade Python OpenAPI client generation version

enzo-crance-statnett:
This PR comes from trying to generate an OpenAPI client from `main`, which actually does not work.

We probably got some updates recently that caused the generated client to be a bit different (although it works the same way). This triggers a linter that complains about union types being represented the "old" way.

This PR just fixes the flow. When we add new resources, we will need to regenerate the client, and I thought this was cleaner to try to fix this in a previous PR, so that we don't have 300 changed files at the same time as some other meaningful changes later.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -5,7 +5,7 @@ hypothesis
 Jinja2
 markdown-callouts
 mkdocs-material
-openapi-python-client==0.21.6
+openapi-python-client==0.26.2
 openpyxl
 pandas
 psycopg[binary,pool]


krilor (changes requested):
Good stuff.
But cannot be merged as is.

It seems like your local is ahead of our decided-upon-version for generating clients. Ref https://github.com/elhub/flex-information-system/blob/main/requirements-dev.txt#L8C1-L8C22

This [particular change came in 0.26](https://github.com/openapi-generators/openapi-python-client/releases/tag/v0.26.0) while we are at 0.21. We have version pinned this dependency specifically to avoid that each of us generate different clients.

Options:

1. Nuke your venv and init again to get  to the current locked version. Abandon this PR.
2. Update requirements dev to the latest version, then nuke venv and init + generate client. Commit changes to PR and we aim to merge.

I propose 2.

PR #345, opened 2025-10-20
chore: anticipate SQL linter complaint on long resource name (FLEX-792)

enzo-crance-statnett:
This PR just adds newlines in the grants, so that we can fit lines containing `flex.service_providing_group_grid_suspension_comment_history` in a later PR without the SQL linter grumbling.


krilor (dismissed):
Good stuff. I feel like we should have done these as part of this PR as well?

https://github.com/elhub/flex-information-system/pull/343/files#diff-6a6dc6edd1a5495e5fa0fdcb59fa282d732ca44785fc60790618d9e0b69a2bdc

PR #348, opened 2025-10-22
feat: add SPG grid suspension comment (FLEX-791)

enzo-crance-statnett:
This PR adds the SPG GS comment resource.

The UI is in the same PR since it is now a very low weight change thanks to https://github.com/elhub/flex-information-system/pull/347.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -3,6 +3,13 @@
 This changelog tracks all changes that are visible in our API.
 
 <!-- markdownlint-disable MD013 -->
+## 27.10.2025
+
+* **Added the _Service Providing Group Grid Suspension Comment_ resource.**
+
+## 08.10.2025
+
+* **Added the _Service Providing Group Grid Suspension_ resource.**
 
 ## 26.09.2025
 


frontend/src/App.tsx
```diff
@@ -1441,3 +1484,29 @@ export const App = () => (
     )}
   </Admin>
 );
+
+// helper components to redirect to previous page after edit/create
```
krilor: NIT: Shouldnt this comment be tied to the actual components to help give context in the IDE?

frontend/src/App.tsx
```diff
@@ -1037,6 +1037,49 @@ export const App = () => (
...
+                    <CommentInput />
+                  </CreateRedirectPreviousPage>
+                </ResourceContextProvider>
+              }
+            />
+            <Route
+              path=":service_providing_group_id/grid_suspension/:service_providing_group_grid_suspension_id/comment/:id"
+              element={
+                <ResourceContextProvider value="service_providing_group_grid_suspension_comment">
+                  <EditRedirectPreviousPage>
```
krilor: Why is this needed here and not on the other resources?

openapi/resources.yml
```diff
@@ -462,6 +462,7 @@ resources:
       - delete
     audit: true
     generate_views: true
+    comments: true
```
krilor: 🥳

frontend/src/App.tsx
enzo-crance-statnett: It comes from the depth level of this resource (SPG -> suspension -> comment). Usually, we have only 2 levels, so we can reconstruct the URL of the redirect by taking the foreign key to the parent resource from `record`, but there we cannot because we don't know the SPG ID (this ID lies in the suspension, not in the comment).

I wonder whether we could make the behaviour uniform, and just use this "previous page redirect" on all edit/create pages defined in this file.

db/flex/service_providing_group_grid_suspension_comment_rls.sql
```diff
@@ -0,0 +1,160 @@
...
+                SELECT
+                    spggsh.impacted_system_operator_id,
+                    spg.service_provider_id
+                FROM
+                    flex.service_providing_group_grid_suspension_history
+                        AS spggsh -- noqa
+                    INNER JOIN flex.service_providing_group AS spg
+                        ON spggsh.service_providing_group_id = spg.id
+                WHERE spggsh.id = service_providing_group_grid_suspension_comment_history.service_providing_group_grid_suspension_id -- noqa
+                UNION ALL
```
krilor: Does it make sense to do "UNION" instead of "UNION ALL" to remove duplicates maybe?

db/flex/service_providing_group_grid_suspension_comment_rls.sql
```diff
@@ -0,0 +1,160 @@
...
+        SELECT 1
+        FROM spggs_history
+            INNER JOIN flex.identity AS comment_creator
+                ON service_providing_group_grid_suspension_comment_history.created_by -- noqa
+                    = comment_creator.id
+        WHERE ((
+            service_providing_group_grid_suspension_comment_history.visibility = 'same_party' -- noqa
+            AND comment_creator.party_id = (SELECT flex.current_party()) -- noqa
+        ) OR (
+            service_providing_group_grid_suspension_comment_history.visibility -- noqa
```
krilor: There is something we have not considered with this visibility stuff. Shouldn't visibility be set according to the _current_ visibility (aka visibility as of now aka the latest visibility). So I create a comment and realize I set the wrong visibility. Shouldn't I be able to correct that?

Esp since it will be there in history for all eternity.

PR #365, opened 2025-10-28
feat: generate RLS policies for comment resources (FLEX-808)

enzo-crance-statnett:
This PR introduces a common template for RLS policies for comment resources. This allows adding new comment resources fast, with high confidence, and with a full consistency between all the comment resources.

This is made possible by introducing a view giving a list of the involved parties in a resource where a comment is attached.

PR #366, opened 2025-10-28
fix: rename parameter in latest visibility function for SPGGSC (FLEX-808)

enzo-crance-statnett:
This PR renames a parameter in the function giving latest visibility for a SPGGS comment.

This change is not strictly necessary before actually using the generation mechanism, but since the former parameter name `id` is also the name of a field in the tables, we got a bug in https://github.com/elhub/flex-information-system/pull/365 because of this. Fixed by renaming + adding a prefix.

> [!NOTE]
> The changes in this PR are temporary anyway, the point is to deploy this hotfix so the automation PR is accepted by Liquibase.

### DIFF
--- a/db/flex/service_providing_group_grid_suspension_comment_rls.sql
+++ b/db/flex/service_providing_group_grid_suspension_comment_rls.sql
@@ -82,7 +82,8 @@ USING (
     )
 );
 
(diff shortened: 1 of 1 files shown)


sigridOC (approved):
Looks good to me :)

PR #370, opened 2025-10-29
feat: add service providing group product suspension resource (FLEX-824)

enzo-crance-statnett:
This PR adds the SPG product suspension resource.


Brynhildur-Eddudottir-Statnett (dismissed):

krilor (changes requested):
Good stuff. I think we need to fix the thing related to the checking the timestamps.

PR #371, opened 2025-10-29
feat: add controllable unit suspension resource (FLEX-821)

sigridOC:
This PR adds the CU suspension resource.

Draft for functional validation before implementing.

### DIFF
--- a/docs/resources/controllable_unit_suspension.md
+++ b/docs/resources/controllable_unit_suspension.md
@@ -83,10 +83,10 @@ No policies.
 
 #### Service Provider
 
-| Policy key | Policy                                                           | Status |
-|------------|------------------------------------------------------------------|--------|
-| CUS-SP001  | Read CU suspensions overlapping with periods where they own the CU            | DONE   |
-| CUS-SP002  | Read history on CU suspensions on periods where they own the CU. | DONE   |
+| Policy key | Policy                                                             | Status |
+|------------|--------------------------------------------------------------------|--------|
+| CUS-SP001  | Read CU suspensions overlapping with periods where they own the CU | DONE   |
+| CUS-SP002  | Read history on CU suspensions on periods where they own the CU.   | DONE   |
 
 #### Third Party
 


openapi/resources.yml
```diff
@@ -168,6 +168,67 @@ resources:
...
+      id:
+        description: Unique surrogate identifier.
+        format: bigint
+        type: integer
+        readOnly: true
+        example: 74
+      controllable_unit_id:
+        description: >-
+          Reference to the controllable unit this relation links to a
+          service provider.
```
enzo-crance-statnett: _Reference to the suspended controllable unit_

docs/resources/controllable_unit_suspension.md
enzo-crance-statnett: As discussed, RLS policies to come. Nothing special to review on this file for now 👍

openapi/resources.yml
```diff
@@ -168,6 +168,59 @@ resources:
...
+        format: bigint
+        type: integer
+        x-no-update: true
+        x-foreign-key:
+          resource: controllable_unit
+          field: id
+        example: 2
+      impacted_system_operator_id:
+        description: >-
+          Reference to the `party` that is the impacted system operator.
```
krilor: I think the description should look more like the one in https://github.com/elhub/flex-information-system/pull/370/files#diff-b08656106e61087cac4596a347d188ee07bf448973803e02c8a44ce49dd93c97

So more like 

> Reference to the impacted system operator suspending the controllable unit.

docs/resources/controllable_unit_suspension.md
```diff
@@ -0,0 +1,107 @@
...
+No policies.
+
+### Field Level Authorization
+
+For party type abbreviations, check [the auth docs](../technical/auth.md#party-market-actors)
+
+| FIELD                       | ANON | BRP | ES | EU | FISO | MO | SO  | SP | TP | ORG |
+|-----------------------------|------|-----|----|----|------|----|-----|----|----|-----|
+| id                          |      | R   | R  | R  | R    | R  | R   | R  | R  |     |
+| controllable_unit_id        |      | R   | R  | R  | RC   | R  | R   | RC | R  |     |
```
krilor: There are a couple of lines that are off here. SP should not have `C` IMHO.

It might be that this is a CSV-in-VSCode-is-awkward-typo . In that case, consider installing a couple extensions.

I have these. The first one is the one I rely the most on.

* https://marketplace.visualstudio.com/items?itemName=mechatroner.rainbow-csv
* https://marketplace.visualstudio.com/items?itemName=janisdd.vscode-edit-csv

docs/resources/controllable_unit_suspension.md
```diff
@@ -0,0 +1,107 @@
...
+| CUS-SO001 | Create, read, update and delete their own CUS.     | TODO   |
+| CUS-SO002 | Read history on their own CUS.                     | TODO   |
+| CUS-SO003 | Read CUS targeted at CUs they can see.            | TODO   |
+| CUS-SO004 | Read history on CUS targeted at CUs they can see. | TODO   |
+
+#### Service Provider
+
+| Policy key  | Policy                                      | Status |
+|-------------|---------------------------------------------|--------|
+| CUS-SP001 | Read CUS targeting their CUs.            | TODO   |
```
krilor: PRO TIP: There is a nice markdown table prettifier extension in VSCode.

* https://marketplace.visualstudio.com/items?itemName=darkriszty.markdown-table-prettify

enzo-crance-statnett:
A few thoughts on what we miss in tests for this PR to be ready. We're not far from there.

SO rules 3 and 4 should be tested by having another SO than the one owning the AP. I propose to have something like an SPG that sends an application to a second SO. So that the SO can see the CU and thereby suspension as well. Then we have a real test for these rules.

SP rules should take into account the case of former SP / former suspension + negative testing (former SP cannot see current suspension, and reverse).

db/flex/controllable_unit_suspension_rls.sql
```diff
@@ -0,0 +1,122 @@
...
+ON controllable_unit_suspension
+FOR SELECT
+TO flex_service_provider
+USING (
+    EXISTS (
+        SELECT 1
+        FROM flex.controllable_unit_service_provider AS cusp
+        WHERE cusp.controllable_unit_id = controllable_unit_suspension.controllable_unit_id -- noqa
+            AND cusp.service_provider_id = (SELECT flex.current_party())
+            AND cusp.valid_time_range @> current_timestamp
```
krilor: Shouldn't this check for overlap between the CUSP valid time and the CUS record time, similar to the history one?

Policy is 

> CUS-SP001 	Read CU suspensions on periods where they own the CU.

The case we are missing here is where SP changes on the CU while the suspension is active. The old SP will loose access.

krilor (changes requested):
I think we need to cahnge the CUS-SP001 implementation

krilor (dismissed):
Nice 🤩

PR #376, opened 2025-10-31
feat: notifications for SPG grid suspension comment (FLEX-795)

enzo-crance-statnett:


backend/event/models.sql
```diff
@@ -318,3 +318,25 @@ WHERE spggsh.service_providing_group_grid_suspension_id = @resource_id
...
+
+-- name: GetServiceProvidingGroupGridSuspensionCommentNotificationRecipients :many
+SELECT DISTINCT
+    unnest(ARRAY[
+        spg.service_provider_id,
+        spggsh.impacted_system_operator_id
+    ])::bigint
+-- using SPGGS(C) history because of visibility + possible deletion
+FROM service_providing_group_grid_suspension_comment_history AS spggsch
+    INNER JOIN service_providing_group_grid_suspension_history AS spggsh
```
krilor: TODO: Evaluate if we can use the `_involved_party` view instead. Requires update of the policy description as well.

PR #382, opened 2025-11-03
feat: SPG product suspension notices (FLEX-826)

enzo-crance-statnett:
Product type no longer qualified + lingering notice

_(This has been validated functionally before implementation.)_

### DIFF
--- a/db/flex/notice.sql
+++ b/db/flex/notice.sql
@@ -375,7 +375,8 @@ CREATE VIEW notice AS (
                         flex.service_providing_group_product_application
                             AS spgpa -- noqa
                     WHERE (
-                        spgpa.prequalified_at IS NOT null
+                        spgpa.status = 'temporary_qualified'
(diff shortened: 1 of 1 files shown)


db/flex/notice.sql
```diff
@@ -355,4 +355,68 @@ CREATE VIEW notice AS (
...
+                    SELECT
+                        spgpa.service_providing_group_id,
+                        spgpa.procuring_system_operator_id,
+                        unnest(spgpa.product_type_ids) AS product_type_id
+                    FROM
+                        flex.service_providing_group_product_application
+                            AS spgpa -- noqa
+                    WHERE (
+                        spgpa.prequalified_at IS NOT null
+                        OR spgpa.verified_at IS NOT null
```
krilor: what about temporary_qualified?

Ref https://github.com/elhub/flex-information-system/pull/370#discussion_r2489028427

PR #386, opened 2025-11-05
docs: conceptual model umbrella

maribjorn:


docs/concepts/conceptual-model.md
```diff
@@ -73,7 +81,24 @@ congestion management.
...
+    Article 2:
+    (21) 'controllable unit' or 'CU' means a single power-generating module and/or demand unit pursuant to Article 2(5) of
+    [RfG NC 2.0] and Article 2(4) of [DC NC 2.0];
+
+    (26) 'service providing group' or 'SPG' means an aggregation of controllable units or service providing units connected
+    to more than one connection agreement point within the same scheduling area. SPG is defined by the service provider
+    to provide balancing or local services;
+  
+    The NC DR also includes a definition of a 'service providing unit' (SPU) as a single unit connected to one connection
+    agreement point. In FIS, a SPU is treated as an SPG where all CUs are connected to the same connection point.
```
krilor: > In FIS, a SPU is treated as an SPG where all CUs are connected to the same connection point.

We do not include SPU in the FIS. We do not see a reason for this distinction and have thus not modelled it in.

docs/diagrams/simplified_model.drawio.png
krilor: A diagram like this one should have a clear language/notation. This diagram has room for improvement in that regard. It should be clear what the different shapes and lines mean. What is a dotted line? Why is there a dotted "resource" around the SO boxes and why are some of those dotted? Why are some lines connected to collections and not specific things? Is that significant? and so on

Other remarks.

* The abbreviation list shouldn't be part of the drawing. Those should be part of the text IMHO.
* EU has no relation to BRP
* We do not use "Product" but "Product type" in the FIS.
* Should we annotate the lines somehow?

docs/concepts/conceptual-model.md
```diff
@@ -0,0 +1,146 @@
...
+| TR           | [Technical Resource](../resources/technical_resource.md)                                            |
+| AP           | [Accounting Point](../resources/accounting_point.md)                                                |
+| EU           | [End User](end-user.md)                                                                             |
+| ES           | [Energy Supplier](../technical/auth.md#energy-supplier)                                             |
+| SP           | [Service Provider](../technical/auth.md#service-provider)                                           |
+| BRP          | [Balance Responsible Party](../technical/auth.md#balance-responsible-party)                         |
+| SO           | [System Operator](../technical/auth.md#system-operator)                                             |
+| PSO          | [Procuring System Operator](../technical/auth.md#system-operator)                                   |
+| CSO          | [Connecting System Operator](../technical/auth.md#system-operator)                                  |
+| CUSP         | [Controllable Unit Service Provider](../resources/controllable_unit_service_provider.md) (contract) |
```
maribjorn: Ok with links like this? I have also included links where concepts are introduced in the text.

docs/concepts/conceptual-model.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Notation
```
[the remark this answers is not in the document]
krilor: I personally like it to be visual and in the diagram. The external notation is somtimes the simplest solution, e.g. when we use generated diagrams. In this case I see no reason for having it as harder-to-read text.
Do note that there are a couple of things that must be corrected .

* (norwegian) spelling
* cusp is not a membership
* what is "external" lookup

docs/concepts/conceptual-model.md
```diff
@@ -0,0 +1,146 @@
...
+    * `1` – exactly one
+    * `1..*` – one or more
+* `blue ellipses` - party type
+* `blue arrows` - dependencies between processes
+    * `solid` - direct party relation to resource
+    * `dotted` - indirect party relation to resource
+
+## What is a controllable unit?
+
+We start introducing the origin of flexibility in the
```
krilor: This sentence does not compute for me. :P

docs/concepts/conceptual-model.md
```diff
@@ -0,0 +1,146 @@
...
+
+> Article 2:
+>
+> (26) 'service providing group' or 'SPG' means an aggregation of controllable
+> units or service providing units connected to more than one connection
+> agreement point within the same scheduling area. SPG is defined by the service
+> provider to provide balancing or local services;
+
+A [*service providing group*](../resources/service_providing_group.md) (SPG) is
+a collection of CUs. The CUs are grouped together within a scheduling area by a
```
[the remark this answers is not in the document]
maribjorn: This is what the network code mentions, and also what Statnett is planning to do. Is it the term "scheduling area" that should be something else, or do you think the CUs could be grouped across areas?

docs/concepts/conceptual-model.md
```diff
@@ -0,0 +1,146 @@
...
+presented elsewhere in the documentation.
+
+## The conceptual model
+
+The simplified conceptual model below illustrates some of the most central
+concepts in the FIS and the relations between them. The model serves as an
+overview when we introduce concepts and abbreviations throughout the documentation.
+
+The conceptual model is based on terminology from the [Network Code on Demand
+Response](../index.md#network-code-on-demand-response-nc-dr) (NC DR) and in the concept
```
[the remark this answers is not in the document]
maribjorn: Adding a reference to the index page.

docs/concepts/conceptual-model.md
maribjorn: I agree! Could it work to just introduce the concepts in the text like I have done now, and remove the table all together? I have mentioned almost all of them already, so could easily include the rest I think.

docs/concepts/conceptual-model.md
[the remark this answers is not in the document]
maribjorn: Added a description of scheduling areas.

PR #387, opened 2025-11-05
feat: play with gh copilot code review custom instructions

krilor:
Inspired by https://github.com/elhub/flex-information-system/pull/382 .
Docs in https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions#creating-path-specific-custom-instructions-2
Example instructions found in https://docs.github.com/en/copilot/how-tos/use-copilot-agents/request-a-code-review/use-code-review#example

### DIFF
--- a/.github/instructions/db.instructions.md
+++ b/.github/instructions/db.instructions.md
@@ -1,0 +1,9 @@
+---
+applyTo: "db/**/*.sql"
+---
+
+# DB Code Review
+
+When performing a code review follow these rules.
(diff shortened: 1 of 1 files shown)


krilor:
> Indeed, I had to manually hide such comments in a previous PR. Good that we have a way to control the way reviews are done.

First try to change the behaviour, so just 🤞that it works. At least it gives us a starting point to modify later on.

PR #388, opened 2025-11-05
docs: notification and notice consept

Brynhildur-Eddudottir-Statnett:


docs/concepts/notification-and-notice.md
```diff
@@ -0,0 +1,159 @@
...
+required from them.
+
+In this case, both the notification and the notice point to the same action that
+is required from the system operator. However, the notice will remain until the
+grid node is set, while the notification is only created once when the CU is
+created.
+
+### Example of data inconsistency notice
+
+Imagine that a service provider (SP) has a current contract with an end user on a
```
maribjorn: Should we introduce the abbreviation here after "service provider" was just introduced above? Same with controllable unit

PR #395, opened 2025-11-06
feat: cusc implementation (FLEX-832)

sigridOC:
This is only a draft as I am missing tests. Once they are added I will change the status of the PR.


backend/data/models/models.go
```diff
@@ -114,7 +114,32 @@ type ApiControllableUnitSuspension struct {
 	RecordedAt               pgtype.Timestamptz
 }
 
-type ApiControllableUnitSuspensionHistory struct {
+type ControllableUnitSuspensionComment struct {
```
krilor: This is not line with the latest changes from #408 . Needs to be aligned with main.

docs/resources/controllable_unit_suspension_comment.md
```diff
@@ -0,0 +1,110 @@
+# Controllable Unit Suspension
```
krilor: There seems to be a copy-pasta error here.

krilor (changes requested):
Lots of good stuff here. Spotted a few things that should be fixed.

test/api_client_tests/test_cus_comment.py
```diff
@@ -0,0 +1,352 @@
...
+    assert not isinstance(u, ErrorMessage)
+
+    # RLS: CUS-FISO002
+    check_history(client_fiso, cus.id)
+
+    # RLS: CUSC-SO002
+    # RLS: CUSC-SP002
+
+
+def test_cus_so(data):
```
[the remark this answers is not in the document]
sigridOC: good spot

PR #404, opened 2025-11-10
feat: notifications for CU suspension (FLEX-822)

enzo-crance-statnett:
This PR adds notifications for CU suspension.

We notify:

- the SP owning the CU at the time of the event
- all PSOs (SO having a SPG product application for an SPG containing the CU at the time of the event)
- all ISOs (only the CSO of the CU as it is our definition of ISO for now).

Event RLS updated and notifications tested locally.
Docs updated to mention PSO because it was in the text but missing from the diagram.


backend/event/models.sql
```diff
@@ -387,3 +387,39 @@ WHERE spgpsch.service_providing_group_product_suspension_comment_id = @resource_
...
+WHERE cush.controllable_unit_suspension_id = @resource_id
+    AND tstzrange(cush.recorded_at, cush.replaced_at, '[]')
+        @> @recorded_at::timestamptz
+    AND tstzrange(cusp.valid_from, cusp.valid_to, '[)')
+        @> @recorded_at::timestamptz
+-- ISO (= CSO)
+UNION
+SELECT ap.system_operator_id
+FROM controllable_unit_suspension_history AS cush
+    INNER JOIN controllable_unit AS cu
```
krilor: IDEA: Should we make a controllable_unit_system_operator view similar to the enduser/energy supplier one? I generally think that we want to have such views for both ap and cu.

One reason to introduce it is because SO is not _really_ a property on a AP. It is (in a fully normalized world) a relationship through the MGA:

```
CU --under--> AP --in--> MGA --owned by--> SO
```

having a view allows us to change that logic independently.

Just thinking out loud

krilor (changes requested):
Requesting changes due to the missing where clause in inline comment.

PR #405, opened 2025-11-10
feat: use ready-for-market functions in event worker

enzo-crance-statnett:
This PR replaces manual ready-for-market type of checks in the event worker with calls to functions already defined in the database and in use for RLS policies / notices.

As the event worker lives outside of the database, it only had access to the `api` schema, so we had to introduce wrapper functions exposing the functions defined in the `flex` schema. This made little sense in the `api` schema and felt like we would mix several use cases in a single schema, so we introduced a new schema.

It is called `notification` to conform to the future plans we had in https://github.com/elhub/flex-information-system/blob/main/docs-dev/modular-monolith.md. This gives us possibility to move stuff away from `api` to `notification` over time as well.

### DIFF
--- a/backend/event/models.sql
+++ b/backend/event/models.sql
@@ -296,10 +296,7 @@ WHERE spggsh.service_providing_group_grid_suspension_id = @resource_id
         @> @recorded_at::timestamptz
     AND tstzrange(spggph.recorded_at, spggph.replaced_at, '[]')
         @> @recorded_at::timestamptz
-    AND (
-        spggph.status IN ('approved', 'conditionally_approved')
-        OR spggph.prequalified_at IS NOT null
-    )
+    AND notification.spg_grid_prequalification_ready_for_market_check(spggph)
 -- PSO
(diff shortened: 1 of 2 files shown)


backend/event/models.sql
krilor: We are now moving away from one schema and should anotate all objects/relations with schema name for clarity/readability.

krilor (approved):
I'm excited to see this and think it is a good idea. Starting to see a more modular database is 🤩

But having objects with the same name in different schemas forces us to be very strict in our use of them. We _MUST_ always use schema prefix to avoid Weird Bugs™️when postgres automagically resolves to the wrong thing.

Eg. things like this:

```sh
❯ rg sp_product_application_ready_for_market_check
db/flex/notice.sql
284:                    WHERE sp_product_application_ready_for_market_check(sppa) -- noqa
```

It is not critical to do prior to merge but we should not consider this work finished until we have schema-prefixed-all-the-things (IMHO)

PR #406, opened 2025-11-10
feat: notices for CU suspension (FLEX-823)

enzo-crance-statnett:


docs/resources/notice.md
```diff
@@ -13,6 +13,7 @@ from the responsible party.
...
 | no.elhub.flex.controllable_unit.grid_validation_status.pending                      | Grid validation status pending                                                          | CSO               | Verify CU for grid verification and update status                                                          |
 | no.elhub.flex.controllable_unit_service_provider.valid_time.outside_contract        | Inconsistency: CUSP valid while end user is not valid on the AP                         | SP                | Update CUSP to match the updated end user data from Elhub                                                  |
+| no.elhub.flex.controllable_unit_suspension.lingering                                | Inactivity: nothing has happened on the suspension in 2 weeks                           | ISO               | Suspension is a temporary procedure. Consider reinstating the CU or unqualifying the SPGs it belongs to.   |
```
krilor:
```suggestion
| no.elhub.flex.controllable_unit_suspension.lingering                                | Inactivity: nothing has happened on the suspension in 2 weeks                           | ISO               | Suspension is a temporary procedure. Consider reinstating the CU and use other lighter mechanisms to limit the use of the CU.   |
```

This is vague on purpose as we don't really have much to go on here. There is no "conditionally_approved" (aka long-term limits) on the CU level. And temporary limits are also erm... temporary ..

But suggesting the use of "unqualifying SPG" is not something we want to do either way. It is a even heavier/strict disabling.

PR #413, opened 2025-11-14
feat: notifications for CU suspension comments (FLEX-833)

enzo-crance-statnett:


sigridOC (dismissed):
Nice one Enzo!
Sorry for the mistakes I had done..

PR #434, opened 2025-11-24
docs: spg bidding zone concept (FLEX-850)

krilor:
Documents the use of the term "bidding zone" on SPG.

### DIFF
--- a/docs/concepts/conceptual-model.md
+++ b/docs/concepts/conceptual-model.md
@@ -118,13 +118,13 @@ are close together and all connected to the same part of the grid, the SPG can
 offer local services such as congestion management. If the members are spread
 out over a larger area, the SPG can maybe only offer balancing services.
 
-Each SPG is resticted to a specific area by the service provider. The provider
+Each SPG is restricted to a specific area by the service provider. The provider
 specifies this when creating the SPG. We use the term `bidding zone`, rather
 than `scheduling area` used in NC DR, for this area.
 [SOGL](https://eur-lex.europa.eu/eli/reg/2017/1485/oj/eng) tells us that the
 geographical scope of the scheduling area is equal to the bidding zone. The term
 `bidding zone` is widely used and easy to understand for market participants.
-The norwegian word for bidding zone is `budområde`, and in common language also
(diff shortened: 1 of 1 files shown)


docs/concepts/conceptual-model.md
```diff
@@ -111,10 +111,28 @@ NC DR defines a service providing group like this:
...
-Norway the scheduling areas are the same as the bidding zones or price areas
-NO1 - NO5.
+a collection of CUs. We call the CUs *members* of the SPG.
+
+The topological location of the CUs limits the use of the SPG. If SPG members
+are close together and all connected to the same part of the grid, the SPG can
+offer local services such as congestion management. If the members are spread
+out over a larger area, the SPG can offer balancing services.
+
+Each SPG is resticted to a specific area by the service provider specifying when
```
maribjorn: resticted -> restricted

docs/concepts/conceptual-model.md
```diff
@@ -111,10 +111,28 @@ NC DR defines a service providing group like this:
...
+creating the SPG. We use the term `bidding zone`, rather than `scheduling area`
+used in NC DR. [SOGL](https://eur-lex.europa.eu/eli/reg/2017/1485/oj/eng) tells
+us that the geographical scope of the scheduling area is equal to the bidding
+zone. The norwegian word for bidding zone is `budområde`, and in common language
+also refrerred to as `prisområde`. We have chosen to use this term because:
+
+* The term 'bidding zone' is already widely used.
+* It is easier to understand for market participants, i.e. service providers.
+* We can strech the definition a bit to cover both local and
+  balancing services/markets.
```
maribjorn: Hmm, I think we have to be careful to use the definitions consistently at least. How can we make it cover local markets, or in what way would we use this definition? Or does the NC DR imply that local markets also will have their own "scheduling areas"/"bidding zones"? That would need a new concept in my opinion.

Scheduling areas are so far defined in SO GL: "‘scheduling area’ means an area within which the TSOs' obligations regarding scheduling apply due to operational or organisational needs;"
And in Norway the bidding zones match these scheduling areas.

maribjorn (approved):
Looks very good to me!

PR #456, opened 2025-12-02
fix: changesets checksum make liquibase fail

krilor:
We have recently removed `notes`. This makes the changeset checksum fail - which makes liquibase sad.

Switching to `runOnChange: true` allows us to not keep track for checksums, and since the table is created `IF NOT EXISTS`, its a safe pattern. I think we do this in other places as well.


enzo-crance-statnett (approved):
We do this in comments indeed. As they are generated, we can't have checksums.

Works here too :+1:

PR #471, opened 2025-12-05
refactor: use event subject (FLEX-902)

enzo-crance-statnett:
This PR makes use of the newly introduced `event.subject` field. Basically, whenever a resource `R` has a parent resource `P`, its events are now encoded with `source = P` and `subject = R`.

The `capture_event` trigger could be kept generic. It keeps taking the source from the argument, but now it infers the subject from the name of the table concerned by the event. If no argument is given, the resource is toplevel in the hierarchy, so we don't need to set a subject and the table name goes to the source.

RLS are maintained and this update allows regrouping several policies into one in several cases.

Screenshot 2025-12-05 154528

### DIFF
--- a/backend/event/models.sql
+++ b/backend/event/models.sql
@@ -239,11 +239,11 @@ FROM api.controllable_unit AS cu
     INNER JOIN notification.controllable_unit_system_operator AS cuso
         ON cu.id = cuso.controllable_unit_id
 WHERE cu.id = (
-    SELECT controllable_unit_id
-    FROM api.technical_resource_history trh
-    WHERE trh.technical_resource_id = @resource_id
-    LIMIT 1
-)
+        SELECT controllable_unit_id
+        FROM api.technical_resource_history trh
+        WHERE trh.technical_resource_id = @resource_id
+        LIMIT 1
+    )
     AND cu.status != 'new'
     AND cuso.valid_time_range @> @recorded_at::timestamptz;
 


db/flex/event.sql
```diff
@@ -30,22 +30,48 @@ LANGUAGE plpgsql
 AS
 $$
 DECLARE
-    id bigint;
-    operation text;
-    event_data jsonb;
+    -- source of the event
+    l_source text;
+    l_src_id bigint;
```
krilor: NIT: should keep consisten naming: `l_source` vs `l_src`

krilor (approved):
Image 🤩

Well done! I think this is a really good improvement on our events and being able to do it with a negative line count is pretty solid.

I've tried my best at reviewing every line of change. Esp the access policies are a bit hard to reason about, but I was not able to spot anything.

krilor:
I've converted this to a draft because we cannot merge this until the event worker code in the backend is updated to handle this new stuff. Ref https://github.com/elhub/flex-information-system/blob/main/backend/event/event.go#L24

One problem i foresee is that our current very bad implementation of pgoutput decoding assumes consistent column ordering of the events table across environments. This change https://github.com/elhub/flex-information-system/pull/470/files#diff-831af93ed525d56c6cadd8fceaf0ae32ee7251adac4b4e48ec9704dd1fcd76f4 does not ensure that. To ensure that "migrated" and "new" environments have the same column ordering we can simply move the new columns (subject_resource, subject_id) last in the CREATE TABLE statement. Basically move them after "recorded_by".

We could of course fix the pgoutput implementation, but this simple workaround ^^ could be good enough for now.

Once the new event worker logic is merged, we are safe to merge this code.

test/api_client_tests/test_event.py
```diff
@@ -430,7 +430,7 @@ def test_event_sp(sts):
         e
         for e in events
         if e.type_ == "no.elhub.flex.technical_resource.create"
-        and e.source == f"/technical_resource/{tr.id}"
+        and e.subject == f"/technical_resource/{tr.id}"
```
s0lvang: Should we assert source in these tests aswell? Or is it covered by another test?

db/flex/event.sql
```diff
@@ -2,7 +2,7 @@
 -- Manually managed file
 
 -- changeset flex:event-create runOnChange:false endDelimiter:--
---validCheckSum: 9:526c626b33806895131a9e3537b5b1d6
+--validCheckSum: 9:d51ff0412749453fa4f79b1db182e4c4
```
krilor: NIT: consider runAlways over maintaining checksums

PR #476, opened 2025-12-09
fix(ui): resource button (FLEX-903)

enzo-crance-statnett:
This PR replaces the former broken resource button with a button that simply redirects to the resource specified in the `source` field of the event/notice.

As the button becomes trivial and there is no free lunch, the logic has to show up somewhere else.

The plan was to associate flat-style URLs, with only one ID and one resource no matter the nesting level (_e.g._, `/service_providing_group_grid_suspension_comment/12`), to the existing components, and possibly adapt them so that they would support both styles. But in nested resources, this caused issues when clicking on rows of subresource lists, other buttons started failing or were in some bugged limbo state, _etc_. This "catch-all" approach is not easy to follow and the code probably becomes harder to maintain eventually because of all the cases we need to think about.

Instead, we _redirect_ the flat-style URLs to the ones we currently use, so that the new URLs don't break the system, we don't need specific logic in the components, and everything continues to work as before. We just have to declare the possible flat URLs we want to support, and add a redirect component to every page we want to be able to reach from a flat URL.

Tested successfully for events. Draft until it is tested for all notice types.

### DIFF
--- a/.teamcity/pom.xml
+++ b/.teamcity/pom.xml
@@ -13,7 +13,7 @@
     </parent>
 
     <properties>
-        <devxp-build-configuration.version>1.52.0</devxp-build-configuration.version>
+        <devxp-build-configuration.version>1.53.7</devxp-build-configuration.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     </properties>
 
--- a/db/flex/accounting_point_bidding_zone.sql
+++ b/db/flex/accounting_point_bidding_zone.sql
@@ -1,0 +1,21 @@
+--liquibase formatted sql
+-- Manually managed file
+
+-- changeset flex:accounting-point-bidding-zone-create runAlways:true endDelimiter:--
+CREATE OR REPLACE VIEW accounting_point_bidding_zone
+WITH (security_invoker = true) AS (
+    SELECT
+        ap_mga.accounting_point_id,
+        mga.price_area AS bidding_zone,
+        ap_mga.valid_time_range
+    FROM flex.accounting_point_metering_grid_area AS ap_mga
+        INNER JOIN flex.metering_grid_area AS mga
+            ON ap_mga.metering_grid_area_id = mga.id
+);
+
+-- changeset flex:accounting-point-bidding-zone-grants runAlways:true endDelimiter:;
+GRANT SELECT ON TABLE accounting_point_bidding_zone
+TO flex_internal_event_notification;
+
+GRANT SELECT ON TABLE accounting_point_bidding_zone
+TO flex_common;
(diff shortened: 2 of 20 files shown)


frontend/src/components/resource.tsx
```diff
@@ -0,0 +1,59 @@
...
+      component={Link}
+      to={`/${resource}/${id}/show`}
+      label="Go to resource"
+      startIcon={<DataObjectIcon />}
+      {...rest}
+    />
+  );
+};
+
+function hierarchyPath(record: any, resource: string): string {
```
krilor: I get the _intent_ of this, but question if we really need this stuff.

I'd argue that having the "flat" resource is something we want, either way. Why do we force the redirect?

Example: service_providing_group_grid_suspension

We generally want to have a route that is `/service_providing_group_grid_suspension` where e.g. SP, SO, FISO can see all their suspension, _regardless of the SPG it is attached to_ - a list page.

When we have a list page, then we naturally should have `/service_providing_group_grid_suspension/{id}`. As a show page. As I user, I'm now not in the "hierarchy", but rather just looking at individual SPGGS.

If I enter the show-page, then I can immediately go the SPG if I want (there is a referencefield/link). I don't need it to show up in the breadcrumbs or automatically redirect me. I have it all available as part of the fields.

**So** - if i get the flat show url `/service_providing_group_grid_suspension/{id}` - just leave me there

**That said** - I guess we still need to have/support navigating to subresources when _being in the hirerarcy_. If I'm at `/service_providing_group/{id}/show` then the button that sends me to SPG GS should send me to `/service_providing_group/{id}/grid_suspension/{id}/show`.

Does this make sense at all or am I missing something crucial here?

s0lvang: I agree. I think redirecting is not the best approach, I think we at need to figure out how resolve the path without redirecting first

enzo-crance-statnett: Here's what breaks when we start using a flat URL. Example is CU suspension:

- As the CUS list page was written at the time CUS only was a subresource, it lacks information in standalone mode. We did not include ID, CU ID, and audit fields for instance. Can be fixed with a hook telling us whether we are in flat mode or not.

- The nested history button cannot work in this context, the logic is strongly tied to the URL. But if we implement the hook above, we can maybe make a dumber button for flat mode, that just appends `_history` to the URL or something.

- Anything related to subresources in the CUS show page. Currently comments. All URLs become broken, because the logic of subresources is to append to the URL as a path, which only works if we go 100% nested. We can't have `/controllable_unit_suspension/12/comment/7`, we need either fully flat or `/controllable_unit/1/suspension/6/comment/19` as I understand it.

Edit button is custom so it's ok, and event button uses resource context so independent of URL so everything works.

All this sounds fixable. Thankfully we just have one resource under CUS and it's the same for other suspensions (they just have comments).

enzo-crance-statnett:
Successfully tested.

We just end up in a broken state when we try to jump to a resource that does no longer exist. I don't know how to fix this in a simple way.

PR #483, opened 2025-12-11
feat: simpler and more general EVENT-COM001 event policy (FLEX-627)

enzo-crance-statnett:
This PR introduces a general event policy where SO/SP can read events whenever they can read a resource. This covers the former policies and simplifies reasoning about access to events.


krilor (approved):
Good stuff! This approach is solid, I would say. I've been trying to think of places where this would not be valid, and I can think of none! Great improvement

s0lvang (approved):

PR #485, opened 2025-12-11
fix(ui): resource button checks if resource exists before redirecting (FLEX-911)

enzo-crance-statnett:
This PR prevents broken redirects from happening on notifications about deleted resources, by checking whether the resource exists before doing the redirect.

### DIFF
--- a/frontend/src/components/ResourceButton.tsx
+++ b/frontend/src/components/ResourceButton.tsx
@@ -25,7 +25,7 @@ export const ResourceButton = (props: ButtonProps & { source: string }) => {
           navigate(`/${resource}/${id}/show`);
         } else {
           notify(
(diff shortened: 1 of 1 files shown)


frontend/src/components/ResourceButton.tsx
```diff
@@ -1,18 +1,35 @@
...
+        // fetch the resource to ensure it still exists before redirecting
+        const { data } = await dataProvider
+          .getOne(resource, { id })
+          .catch(() => ({ data: undefined })); // not found
+
+        if (data) {
+          navigate(`/${resource}/${id}/show`);
+        } else {
+          notify(
+            `The concerned resource (${resource}/${id}) no longer exists.`,
```
krilor: Another scenario is that the user have lost access to it. Actually probably the most likely scenario.

PR #490, opened 2025-12-15
feat(ui): internationalisation of resource fields in list/show pages (FLEX-913)

enzo-crance-statnett:
This PR uses the field labels introduced recently to render individual fields in the various resources. To this point, only show and list pages are concerned. Comment pages and history list pages should be updated as well.

An important change is the _transposition_ of the field labels generated file. It is easier to just select a big dictionary with only one language, than having to reselect the language everytime. This way, we just change this dictionary when the language is changed.

For now, we use React-Admin's internationalisation capabilities, but as we rely on a custom file and logic in the i18n provider, it will be easy to phase out in order to have a fully custom implementation later.

I tried to infer the keys from current resource in the context + current field (`source`), but the `Labeled` component that is necessary for showing the labels on show pages strictly requires that the _direct_ child have a `label` property, so we are kind of forced to set it manually. Allows for some customisation in more complicated cases though, like some reference fields.

History pages and comments need to be addressed, but this may be merged before.
(incomplete internationalisation does not hurt, stuff like buttons/custom text is not concerned yet anyway)


frontend/src/intl/field-labels.json
```diff
@@ -1,664 +1,695 @@
...
+            "ramp_rate": "Ramp rate",
+            "accounting_point_id": "Accounting point ID",
+            "grid_node_id": "Grid node ID",
+            "grid_validation_status": "Grid validation status",
+            "grid_validation_notes": "Grid validation notes",
+            "validated_at": "Validated at",
+            "recorded_at": "Recorded at",
+            "recorded_by": "Recorded by"
+        },
+        "controllable_unit_suspension": {
```
s0lvang: I would prefer if the dicionary was completly flat, so the keys are "controllable_unit_suspension.id"  since its easier to infer the key-type.

s0lvang (dismissed):
Nice!

I dont think its a huge problem that you have to write the keys of the translation, i prefer flexibility over automation. 

I guess the keys are prefixed with `field` so other translations can be distinguished with something else?

PR #493, opened 2025-12-16
feat(ui): internationalisation of enum fields (FLEX-913)

enzo-crance-statnett:
This PR introduces internationalisation for enum values in the portal, so we don't show raw values but a more human name that goes through the common i18n provider and not hardcoding here and there.

### DIFF
--- a/frontend/src/intl/intl.tsx
+++ b/frontend/src/intl/intl.tsx
@@ -27,11 +27,14 @@ export const useI18nProvider = () => {
   return {
     getLocales: () => [
       { locale: "en", name: "English" },
-      { locale: "nb", name: "Norsk Bokmål" },
-      { locale: "nn", name: "Norsk Nynorsk" },
+      // TODO: uncomment to show language button
+      // { locale: "nb", name: "Norsk Bokmål" },
+      // { locale: "nn", name: "Norsk Nynorsk" },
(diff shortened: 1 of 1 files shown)


enzo-crance-statnett:
Had to declare a few more enums that were not formally enums in the OpenAPI spec.

The zod file changes were just automatically made. Same for test models.

Many files changed beause we have enums a bit everywhere (status f.eks).

text.ts is for some custom text I needed in SPG list header and the role component. Future-proof addition because I guess all the custom text of the app will go there later when we do full internationalisation.

enzo-crance-statnett:
CU lookup, assume party page, and some notice pages are still not completely done. Basically I focused first on the automated/standard resource pages.

frontend/src/components/enum.tsx
```diff
@@ -0,0 +1,59 @@
...
+  useI18nProvider,
+  useRecordContext,
+  useTranslate,
+} from "react-admin";
+import { I18nProvider } from "../intl/intl";
+
+export type EnumFieldProps = {
+  source: string;
+  label?: string; // used only for the header label in a Datagrid
+  enumKey: string;
```
krilor: Future improvement suggestion: make enumKey a uniontype for some safety?

PR #495, opened 2025-12-17
feat: restructure cu show page (FLEX-875)

s0lvang:
image


frontend/src/controllable_unit/balance_responsible_party/ControllableUnitBalanceResponsiblePartyList.tsx
```diff
@@ -1,18 +1,29 @@
...
+import { ControllableUnit } from "../../generated-client";
+import { useParams } from "react-router-dom";
 
 export const ControllableUnitBalanceResponsiblePartyList = () => {
   // accounting point id of the controllable unit whose BRPs we want to get
-  const { accounting_point_id } = useRecordContext()!;
+  const { controllable_unit_id } = useParams<{
+    controllable_unit_id: string;
+  }>();
+  const { data: cu, isLoading } = useGetOne<ControllableUnit & { id: number }>(
```
enzo-crance-statnett: A bit surprised `id` is not included in the `ControllableUnit` type?

s0lvang: Its not required for some reason. Probably because of create

krilor: Ref https://github.com/elhub/flex-information-system/blob/main/docs/technical/openapi-schemas.md

The correct type to use in this case would be the `[resource]_response` aka `controllable_unit_response` aka `ControllableUnitResponse`.

Does that help? If not: where does it go wrong? And can we fix it?

frontend/src/controllable_unit/balance_responsible_party/ControllableUnitBalanceResponsiblePartyList.tsx
```diff
@@ -22,45 +33,53 @@ export const ControllableUnitBalanceResponsiblePartyList = () => {
     "read",
   );
 
+  if (isLoading) {
+    return <Loading />;
+  }
+
+  if (!canRead) {
+    return null;
```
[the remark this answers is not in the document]
krilor: This comment also apply to CUSP

frontend/src/controllable_unit/show/components/Connections.tsx
```diff
@@ -0,0 +1,81 @@
...
+            serviceProvider
+              ? `${serviceProvider?.name} (${serviceProviderRange})`
+              : "No service provider"
+          }
+        />
+        <Link
+          to={`/controllable_unit/${controllableUnit.id}/service_provider`}
+          component={RouterLink}
+        >
+          See previous contracts
```
krilor: You talked about changing the text of this thing. Its not just "previous"

frontend/src/controllable_unit/show/components/LabelValue.tsx
krilor: This component seems misplaced. This is a general component, right? Not just scoped to controllable unit?

frontend/src/controllable_unit/show/components/MetaInfo.tsx
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  const formattedDate = new Date(controllableUnit?.recorded_at).toLocaleString(
+    "no-NB",
+  );
+
+  return (
+    <Stack direction="row" spacing={1}>
+      <Typography color="text.secondary">
+        {translateField("controllable_unit.recorded_at")}: {formattedDate}
```
krilor: Why are we not using the LabelValue thingy here?

frontend/src/controllable_unit/show/useControllableUnitViewModel.ts
```diff
@@ -0,0 +1,200 @@
...
+const getControllableUnitData = async (
+  controllableUnit: ControllableUnit | undefined,
+): Promise<ControllableUnitShowViewModel> => {
+  const controllableUnitIdInt = controllableUnit?.id ?? 0;
+
+  if (!controllableUnit) {
+    throw new Error("Controllable unit not found");
+  }
+  const technicalResourcesPromise = listTechnicalResource({
+    query: { controllable_unit_id: "eqqq." + controllableUnitIdInt },
```
krilor: BUG: 'eqqq.'

krilor (changes requested):
This is a good starting point 👍

I added a few inline comments. 

Requesting changes due to the comment marked as BUG. It probably breaks the frontend, so cannot merge that. Consider if some of the other comments are worth adressing before new review as well.

frontend/src/controllable_unit/show/useControllableUnitViewModel.ts
s0lvang: oof. I was testing the error state. Nice spot

frontend/src/controllable_unit/show/components/MetaInfo.tsx
```diff
@@ -0,0 +1,27 @@
...
+  controllableUnit: ControllableUnit | undefined;
+}) => {
+  const translateField = useTranslateField();
+
+  if (!controllableUnit?.recorded_at) {
+    return null;
+  }
+
+  const formattedDate = new Date(controllableUnit?.recorded_at).toLocaleString(
+    "no-NB",
```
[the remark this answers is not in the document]
s0lvang: Nope should be in both!

frontend/src/controllable_unit/show/ControllableUnitShow.tsx
```diff
@@ -49,10 +49,10 @@ const Layout = () => {
         />
 
         <Connections controllableUnitViewModel={controllableUnitViewModel} />
-        <TechnicalResources
-          controllableUnitId={controllableUnitViewModel.controllableUnit.id}
-          technicalResources={controllableUnitViewModel.technicalResources}
-        />
+
+        <Box sx={{ maxWidth: "1000px" }}>
+          <TechnicalResourceList />
```
krilor: If we are moving back to this way of doing it, we probably want to avoid resolving technical resources in the view thingy as well? Not something we need to deal with as part of this PR, but we should avoid doing it twice.

frontend/src/controllable_unit/technical_resource/TechnicalResourceList.tsx
```diff
@@ -51,9 +52,12 @@ export const TechnicalResourceList = () => {
   return (
     canRead && (
       <ResourceContextProvider value="technical_resource">
+        <Typography variant="h6" sx={{ fontWeight: "bold" }}>
+          Technical resources:
```
krilor: I'm not sure that it is a good idea to put headers like this as part of the List component(s). I commented earlier about this on CUSP.

One of the main reasons I think we should avoid it is the potential re-use of the lists. Some place we want to have it just as a list (e.g. when looking at TR across the whole system).

krilor (approved):

PR #501, opened 2025-12-18
feat(ui): field labels on lookup pages (FLEX-913)

enzo-crance-statnett:


krilor (approved):
One thing we should reconsider is if we really just want to treat these things as "text". Why are these not fields?

We could define `x-intl` here as well: https://github.com/elhub/flex-information-system/blob/main/openapi/openapi-api-base.yml#L88 and read those while generating field names.

PR #503, opened 2025-12-19
feat: file-based env.js (FLEX-928)

krilor:


krilor:
This is a draft since this is a breaking change on how to use the docker image and I need to check if that is ok/do changes in dependent places.

local/frontend/env.test.js
```diff
@@ -0,0 +1,5 @@
+/*<![CDATA[*/
```
enzo-crance-statnett: What is this comment used for?

frontend/Dockerfile
```diff
@@ -1,4 +1,36 @@
...
+
+# We first create a base image for serving static files with nginx.
+# It is intended to be used in scenarios where we want:
+#
+#  * run unprivileged - as nobody
+#  * run with read only root filesystem
+FROM docker.jfrog.elhub.cloud/nginxinc/nginx-unprivileged:stable-alpine-slim AS nginx-unpriv
+
+# We want to run as the nobody user
+ARG UID=65534
```
krilor: NOTE: This requires the container to run with this UID/GID rather than 101 (nginx).

This is based on https://github.com/nginx/docker-nginx-unprivileged/issues/149 and running as "nobody" is quite common in k8s settings.

enzo-crance-statnett (approved):

PR #511, opened 2026-01-05
feat: allow SP to create/update grid node on CU (FLEX-896)

enzo-crance-statnett:
NB: update allowed because they may make a mistake


krilor (changes requested):
I think we should NOT allow update. The field should be controlled by SO, and allowing updates removes that control.

If a mistake has been made then that has to be catched and corrected by SO IMHO. They are still "responsible" for doing the mapping MPID -> grid node.

krilor (approved):

PR #513, opened 2026-01-05
refactor(deploy): remove users and structure data from Liquibase changelogs (FLEX-925)

enzo-crance-statnett:
This PR removes users and structure data from the existing Liquibase changelogs. The goal is to move that to a separate changelog that will be triggered by a distinct Ansible task when we want to load this data.

cf sister PR https://github.com/elhub/flex/pull/602


krilor (changes requested):
Additional changes that should be part of this PR.

1. Removing structure data files to properly clean up
2. Moving users.sql to a location more related to our local dev environment (maybe somewhere in `local` ?)
3. Running users.sql as part of load (somewhere in the just recipe?)

I fear that w/o doing 3 we are breaking our local setup cuz there will be no users.

krilor (approved):

PR #514, opened 2026-01-06
feat: add dialog for activating controllable unit (FLEX-931)

s0lvang:


krilor (approved):
Assuming that you will add a why-comment ref the Copilot thingy: approved

PR #521, opened 2026-01-07
fix: support events with no subject ID in pgoutput (FLEX-924)

enzo-crance-statnett:
Aims to fix
```
panic: runtime error: index out of range [8] with length 7 [recovered, repanicked]
```

### DIFF
--- a/backend/event/event.go
+++ b/backend/event/event.go
@@ -67,10 +67,8 @@ func fromTupleData(td *pgoutput.TupleData) (event, error) {
 	// We don't need it since we can infer it from `type`
 
 	// try to parse subject ID (it can be null if the event concerns a toplevel resource)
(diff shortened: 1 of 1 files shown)


backend/event/event.go
```diff
@@ -67,7 +67,12 @@ func fromTupleData(td *pgoutput.TupleData) (event, error) {
 	// We don't need it since we can infer it from `type`
 
 	// try to parse subject ID (it can be null if the event concerns a toplevel resource)
-	subjectIDStr := string(td.Columns[8])
+	var subjectIDStr string
```
krilor: Nice!

BUT: This implementation is getting worse and worse :P We really should fix this properly at some point (not not in the go codebase)

krilor (dismissed):

PR #532, opened 2026-01-14
refactor: dispatch time-dependent MGA data into internal resources and hide them from the API (FLEX-949)

enzo-crance-statnett:
This PR copies data from MGA to two new internal tables MGA-SO and MGA-PA. The data is hidden from MGA's resource API, but the AP-SO and AP-BZ views are repointed to these tables behind the scenes, so the information is surfaced in the API from the new tables, for now in AP.SO and AP-BZ. Test data got updated so that initial data is filled correctly there.

We now need another PR to actually remove the extra fields from the MGA table internally, since they are no longer used, and we need to update the structure data in the other repo so it reflects the changes here.

### DIFF
--- a/db/api/grants/field_level_authorization.sql
+++ b/db/api/grants/field_level_authorization.sql
@@ -5268,6 +5268,7 @@ TO flex_third_party;
 GRANT SELECT (
     id,
     business_id,
+    business_id_type,
     name
 ) ON TABLE
 api.metering_grid_area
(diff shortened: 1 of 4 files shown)


enzo-crance-statnett:
> [!NOTE]
> We mutate the MGA view freely in this PR because we know 56f28383eab04ddd8d5114d7b4e17e5c42d8cf0a has not been deployed. Liquibase works correctly on these changes when run from a state before that commit, to make it one migration step covering both the "wrong" changes there and the corrections of the current PR.

krilor (approved):
Good stuff :)

I think its a good strategy to do cleanup in a follow-up PR 👍

PR #533, opened 2026-01-14
feat: generate tooltips with intl script and add tooltips to cu-show (FLEX-942)

s0lvang:
I moved the generation of tooltips to resources_to_intl.py so they generate the same keys for the objects. That way we can use the same type for the keys of Tooltip and Translation. 

Right now the frontend will fail if something has a description but not a translation, which is not ideal. But it works for now. 

This also prepares us for translating descriptions when that time comes :D


enzo-crance-statnett (approved):
Nice!

I kept thinking about tooltips when introducing intl, it sounded like it belonged to the intl part as well. So I think we are quite aligned. I just thought translating them all was too much of an effort for the scope of the task. Future work 😄

PR #537, opened 2026-01-16
feat: party list and show with EDS (FLEX-954)

s0lvang:
Monster-PR as usual. 

- Write adapter components between RA-core and EDS
- implement it in PartyList and PartyShow.
- export every component we use from EDS from components/ui
- Create eslint-rule that disallows direct imports from eds

### DIFF
--- a/frontend/src/party/membership/PartyMembershipList.tsx
+++ b/frontend/src/party/membership/PartyMembershipList.tsx
@@ -5,7 +5,10 @@ import {
   useRecordContext,
 } from "ra-core";
 import { List, Datagrid } from "../../components/EDS-ra/list";
-import { DeleteButton } from "../../components/EDS-ra/buttons";
+import {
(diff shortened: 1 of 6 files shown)


frontend/src/party/membership/PartyMembershipList.tsx
```diff
@@ -1,58 +1,70 @@
...
+
+import { useNavigate } from "react-router-dom";
+import {
+  DateField,
+  IdentityField,
+  ReferenceField,
+  TextField,
+} from "../../components/EDS-ra/fields/index";
 import { Permissions } from "../../auth/permissions";
+import { IconPlus, IconSearch } from "@elhub/ds-icons";
```
krilor: Not sure if it makes sense, but should we import this via `components/ui/icons`? Where we `export * from "@elhub/ds-icons"` - if that is even possible.

PR #545, opened 2026-01-21
feat(db): synced persistent notices (FLEX-964)

enzo-crance-statnett:
This PR introduces a `notice` _table_ and temporarily redirects the API view to the `notice_fresh` view, that is exactly the former `notice` view that got renamed (because it is currently used to identify fresh notices), so that no change is visible from the API.

We can now freely work on the logic filling/emptying the notice table, then we will be able to redirect the view to `flex.notice` when we are ready, and possibly delete the `notice_fresh` view.

### DIFF
--- a/db/flex/notice.sql
+++ b/db/flex/notice.sql
@@ -33,11 +33,9 @@ CREATE TABLE IF NOT EXISTS notice (
     )
 );
 
--- changeset flex:notice-index runAlways:true endDelimiter:;
-CREATE INDEX IF NOT EXISTS notice_index_party_id
-ON flex.notice (party_id);
-CREATE INDEX IF NOT EXISTS notice_index_status
-ON flex.notice (status);
+-- changeset flex:notice-index runAlways:true endDelimiter:--
+CREATE INDEX IF NOT EXISTS notice_index_party_id_status
+ON flex.notice (party_id, status);
 
 -- changeset flex:notice-capture-event runOnChange:true endDelimiter:--
 CREATE OR REPLACE TRIGGER notice_event


db/flex/notice.sql
```diff
@@ -1,468 +1,26 @@
...
-                    = qpts.system_operator_id
-        WHERE NOT spps.product_type_ids <@ coalesce(qpts.product_type_ids, '{}')
+-- changeset flex:notice-create runAlways:true endDelimiter:--
+CREATE TABLE IF NOT EXISTS notice (
+    id bigint PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+    party_id bigint NOT NULL,
+    source text NOT NULL,
+    type ltree NOT NULL,
+    data jsonb,
+    recorded_at timestamptz NOT NULL DEFAULT localtimestamp,
```
enzo-crance-statnett: > [!NOTE]
> No proper audit proposed here because:
> - we won't modify notices
> - they will always be added as the system (`recorded_by` is useless)
> - they will need to disappear when they are solved
> - no need for notice history?

Trying to keep things simple until it is proved that the complexity is needed.

db/flex/notice_fresh.sql
```diff
@@ -0,0 +1,207 @@
...
+    UNION ALL
+    SELECT * FROM notice_party_outdated
+    UNION ALL
+    SELECT * FROM notice_party_residual
+);
+-- noqa: enable=AM04
+
+-- changeset flex:notice-sync-function runOnChange:true endDelimiter:--
+-- synchronise the notice table with the fresh notice discovery
+CREATE OR REPLACE FUNCTION notice_sync()
```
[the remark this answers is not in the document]
enzo-crance-statnett: > [!CAUTION]
> Extra careful read of the documentation and comparison of docs for [v16](https://www.postgresql.org/docs/16/sql-merge.html#:~:text=Conversely%2C%20if%20the%20WHEN%20clause%20specifies%20WHEN%20NOT%20MATCHED%20and%20the%20candidate%20change%20row%20does%20not%20match%20a%20row%20in%20the%20target%20table%2C%20the%20WHEN%20clause%20is%20executed%20if%20the%20condition%20is%20absent%20or%20it%20evaluates%20to%20true.) and [v17](https://www.postgresql.org/docs/17/sql-merge.html#:~:text=to%20true.-,If%20the%20WHEN%20clause%20specifies%20WHEN%20NOT%20MATCHED%20BY%20SOURCE%20and,if%20the%20condition%20is%20absent%20or%20it%20evaluates%20to%20true.,-condition) show that the comments above are exactly the opposite of the reality in v16. See below.

The `NOT MATCHED` in previous versions has implicit `BY TARGET`, it is the `BY SOURCE` that does not exist before v17. This means that we _can_ include the `INSERT` statement in the `MERGE`, and the thing we cannot include right now is the update that sets notices not recomputed to `resolved`.

Therefore, we have 2 statements in the latest solution, one `MERGE` containing most of the logic, and one little `UPDATE` that will disappear as soon as we upgrade to v17 🙂

db/flex/notice.sql
```diff
@@ -33,6 +33,12 @@ CREATE TABLE IF NOT EXISTS notice (
     )
 );
 
+-- changeset flex:notice-index runAlways:true endDelimiter:;
+CREATE INDEX IF NOT EXISTS notice_index_party_id
```
krilor: Shouldn't we do a composite index rather than two distinct ones? At least current docs states it as "an index". What did we consider when deciding 1 vs 2 indices?

Either we should bring doc and implementation in sync

krilor (approved):
🤩

PR #710, opened 2026-03-25
feat: align CU show page with SPG show page pattern (FLEX-1107)

s0lvang:
## Summary

- Rewrites `ControllableUnitShow` as a standalone component (drops `Show`/`ShowBase` wrapper) using the same `useQuery` + two-column grid pattern as `ServiceProvidingGroupShow`
- Extracts a new `ControllableUnitShowSummary` panel (id, business_id, accounting point, service provider, BRP, max active power, regulation direction) and a `ControllableUnitShowTabs` component with a "Technical resources" tab ready for future tabs
- Migrates `ActivateControllableUnitButton` from MUI `Button` to EDS `Button`; fixes react-admin `Link` import in `ControllableUnitAlerts`
- Removes the now-obsolete `Header.tsx`, `Connections.tsx`, and `TechnicalInformation.tsx` component files

Closes FLEX-1107


frontend/src/controllable_unit/show/ControllableUnitShow.tsx
```diff
@@ -1,78 +1,101 @@
...
-};
+  if (!cu || !viewModel) {
+    return null;
+  }
 
-export const ControllableUnitShow = () => {
-  const { id } = useParams<{ id: string }>();
+  const canUpdateControllableUnit =
+    !!permissions?.allow("controllable_unit", "update") &&
+    (viewModel.technicalResources?.length ?? 0) > 0;
```
enzo-crance-statnett, line 71: This should be called `canActivateControllableUnit` because it is possible to generally update the CU even though there are no technical resources.

frontend/src/controllable_unit/show/ControllableUnitShow.tsx
```diff
@@ -1,78 +1,101 @@
...
+  {
+    status:
+      | "ongoing"
+      | "failed"
+      | "approved-with-warning"
+      | "approved"
+      | "stopped"
+      | "temporarily-stopped"
+      | "pending"
+      | "rejected";
```
enzo-crance-statnett, line 33: I guess these are the accepted status values in the `Badge` component? Is it possible to refer to such an enum? (since we don't use them all below it would be a bit more explanatory)

enzo-crance-statnett (dismissed):

frontend/src/controllable_unit/show/ControllableUnitShow.tsx
s0lvang: Yeah agree!

frontend/src/controllable_unit/show/ControllableUnitShow.tsx
s0lvang: The enum is not exposed from EDS sadly

PR #721, opened 2026-04-07
feat: Synchronization of accounting point data (FLEX-1064)

kenneth-strand-statnett:


kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointRepository.kt
```diff
@@ -115,29 +196,262 @@ class AccountingPointRepositoryImpl : AccountingPointRepository {
...
+                        // Update changed records
+                        conn.prepareStatement(UPDATE_CHANGED_AP_ENERGY_SUPPLIER).use { stmt ->
+                            stmt.setInt(1, energySupplierPartyId)
+                            stmt.setTimestamp(2, validFrom)
+                            stmt.setTimestamp(3, validTo)
+                            stmt.setInt(4, accountingPointId)
+                            stmt.setTimestamp(5, validFrom)
+                            stmt.setInt(6, energySupplierPartyId)
+                            stmt.setTimestamp(7, validTo)
+                            stmt.execute()
```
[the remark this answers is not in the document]
kenneth-strand-statnett: Fixed

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointRepository.kt
```diff
@@ -115,29 +196,262 @@ class AccountingPointRepositoryImpl : AccountingPointRepository {
...
+
+                        // Step 2 — update changed records (same start, different end_user or valid_to).
+                        conn.prepareStatement(UPDATE_CHANGED_AP_END_USER).use { stmt ->
+                            stmt.setInt(1, endUserPartyId)
+                            stmt.setTimestamp(2, validFrom)
+                            stmt.setTimestamp(3, validTo)
+                            stmt.setInt(4, accountingPointId)
+                            stmt.setTimestamp(5, validFrom)
+                            stmt.setInt(6, endUserPartyId)
+                            stmt.setTimestamp(7, validTo)
```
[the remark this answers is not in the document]
kenneth-strand-statnett: Fixed

local/accounting-point/README.md
```diff
@@ -9,4 +9,4 @@ To spin it up, run the following command in the root of the repository:
 ```
 
 You can then access the API at
-[http://127.0.0.1:8080/accounting_point/133700000000020068](http://127.0.0.1:8080/accounting_point/133700000000020068)
+[http://127.0.0.1:5001/accounting_point/133700000000020068](http://127.0.0.1:5001/accounting_point/133700000000020068)
```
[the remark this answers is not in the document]
kenneth-strand-statnett: This is correct

db/flex/accounting_point_rls.sql
```diff
@@ -28,3 +28,12 @@ ON accounting_point
 FOR SELECT
 TO flex_common
 USING (true);
+
+GRANT INSERT, SELECT ON accounting_point TO flex_internal_data;
```
krilor: I wonder if we are missing update/delete here?

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointQuery.kt
```diff
@@ -0,0 +1,101 @@
...
+val CURRENT_CONTROLLABLE_UNIT_ACCOUNTING_POINT = """
+    SELECT accounting_point_id::bigint, accounting_point_business_id::text
+    FROM api.current_controllable_unit_accounting_point(?::text)
+""".trimIndent()
+
+const val GET_ACCOUNTING_POINT_BY_BUSINESS_ID =
+    "SELECT id, business_id FROM flex.accounting_point WHERE business_id = ?"
+
+const val GET_ACCOUNTING_POINT_BY_BUSINESS_ID_FOR_UPDATE_SKIP_LOCKED =
+    "SELECT id, business_id FROM flex.accounting_point WHERE business_id = ? FOR UPDATE SKIP LOCKED"
```
krilor: Shouldn't this be on accounting_point_sync?

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointQuery.kt
```diff
@@ -0,0 +1,101 @@
...
+    "INSERT INTO flex.entity (name, type, business_id, business_id_type) VALUES (?, ?, ?, ?) ON CONFLICT (business_id) DO NOTHING"
+
+const val SELECT_END_USER_PARTY_BY_ENTITY =
+    "SELECT id FROM flex.party WHERE entity_id = ? AND type = 'end_user'"
+
+val INSERT_END_USER_PARTY = """
+    INSERT INTO flex.party (entity_id, name, type, role)
+    SELECT ?, ?, 'end_user', 'flex_end_user'
+    WHERE NOT EXISTS (
+        SELECT 1 FROM flex.party WHERE entity_id = ? AND type = 'end_user'
```
krilor: I think we should enforce this constraint (only one end_user per entity_id) as an actual  constraint and use on CONFLICT DO NOTHING. This impementation is open for race conditions (i think)

If we need to return the ID, then we can follow a pattern like this: https://stackoverflow.com/a/42217872

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointQuery.kt
```diff
@@ -0,0 +1,101 @@
...
+        valid_time_range = tstzrange(?::timestamptz, ?::timestamptz, '[)')
+    WHERE accounting_point_id = ?
+    AND lower(valid_time_range) = ?::timestamptz
+    AND (
+        end_user_id IS DISTINCT FROM ?
+        OR upper(valid_time_range) IS DISTINCT FROM ?::timestamptz
+    )
+""".trimIndent()
+
+val INSERT_NEW_AP_END_USER = """
```
krilor: Why are we using UPDATE+INSERT instead of merge? We want as few roundtrips to the database as possible and using merge will save a few (esp in PG17)

krilor (commented):
I've only been able to look at a bit of this. Added some inline comments. Also, some high-level thoughts/questions

1. Did we evaluate sqlc?
2. We are not setting any lock timeouts - we should probably do that
3. Shouldn't this include some docs updates? An implementation like this should maybe have an accompanying diagram (graphviz/dot) for future us to understand what is going on?

db/flex/accounting_point_rls.sql
kenneth-strand-statnett: It is sufficient for the current needs, but I can add it :+1:

kbackend/src/test/kotlin/no/elhub/flex/db/FlexTransactionTest.kt
```diff
@@ -0,0 +1,148 @@
...
+                        // Nested call — joins the outer transaction.
+                        with(internalPrincipal) {
+                            flexTransaction { _ ->
+                                outerConn.prepareStatement(
+                                    "INSERT INTO flex.accounting_point (business_id) VALUES (?)"
+                                ).use { stmt ->
+                                    stmt.setString(1, businessId)
+                                    stmt.execute()
+                                }
+                                Unit.right()
```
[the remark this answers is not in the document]
kenneth-strand-statnett: Fixed

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointQuery.kt
kenneth-strand-statnett: Fixed

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointQuery.kt
kenneth-strand-statnett: We'd need to add a unique constraint on `accounting_point_id, lower(valid_time_range)` to have a `ON CONFLICT` statement. Alternatively we could use a query with two logical operations(CTE update + insert) What do you think?

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointRepository.kt
```diff
@@ -115,29 +196,262 @@ class AccountingPointRepositoryImpl : AccountingPointRepository {
...
+            }
+        }
+}
+
+/**
+ * Builds a PostgreSQL text array from a list of Instants representing valid_from values.
+ * Used for the `!= ALL(?::timestamptz[])` clause in the stale-record DELETE.
+ */
+private fun Connection.createTimestampArray(instants: List<Instant>): Array =
+    createArrayOf("text", instants.map { it.toSqlTimestamp().toString() }.toTypedArray())
```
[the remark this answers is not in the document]
kenneth-strand-statnett: Fixed

kenneth-strand-statnett:
> I've only been able to look at a bit of this. Added some inline comments. Also, some high-level thoughts/questions
> 
> 1. Did we evaluate sqlc?
> 2. We are not setting any lock timeouts - we should probably do that
> 3. Shouldn't this include some docs updates? An implementation like this should maybe have an accompanying diagram (graphviz/dot) for future us to understand what is going on?

1. Enzo started to look at sqlc
2. Fixed
3. Do we need anything in addition to the ones already made? Ref https://github.com/elhub/flex-information-system/compare/main...FLEX-1068-ap-sync-logic

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointQuery.kt
kenneth-strand-statnett: Fixed

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointQuery.kt
```diff
@@ -0,0 +1,119 @@
...
+val UPSERT_ACCOUNTING_POINT = """
+    WITH ins AS (
+        INSERT INTO flex.accounting_point (business_id)
+        VALUES (?)
+        ON CONFLICT (business_id) DO NOTHING
+        RETURNING id
+    )
+    SELECT id FROM ins
+    UNION ALL
+    SELECT id FROM flex.accounting_point WHERE business_id = ? AND NOT EXISTS (SELECT 1 FROM ins)
```
enzo-crance-statnett: We can just `SELECT id FROM flex.accounting_point WHERE business_id = ?` after the CTE here. Every stateful CTE is evaluated whether it's used or not.
cf [docs](https://www.postgresql.org/docs/current/queries-with.html#:~:text=Data%2Dmodifying%20statements%20in%20WITH%20are%20executed%20exactly%20once%2C%20and%20always%20to%20completion%2C%20independently%20of%20whether%20the%20primary%20query%20reads%20all%20(or%20indeed%20any)%20of%20their%20output.)

Also the `UPSERT` name is weird, we only do `INSERT` here. BTW the sequence diagram should be updated because we never update the AP record. (since it basically only contains a business ID which will never change). Step 10 of the sequence diagram should be deleted.

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/AccountingPointService.kt
```diff
@@ -71,21 +78,70 @@ class AccountingPointServiceImpl(
...
+                                .mapLeft { err -> err.toInternalServerError("lockSyncRowAndMarkStart") }.bind()
+
+                            val endUsers = adapterAccountingPoint.toAccountingPointEndUsers(accountingPointId)
+                            val energySuppliers = adapterAccountingPoint.toAccountingPointEnergySuppliers(accountingPointId)
+
+                            accountingPointRepository.upsertAccountingPointEndUsers(endUsers)
+                                .mapLeft { it.toInternalServerError("upsertAccountingPointEndUsers") }.bind()
+
+                            accountingPointRepository.upsertAccountingPointEnergySupplier(energySuppliers)
+                                .mapLeft { it.toInternalServerError("upsertAccountingPointEnergySupplier") }.bind()
```
enzo-crance-statnett, line 105: As commented already, both of these operations can probably be implemented with a `MERGE` operation similar to what we've done in the [ESBR staging synchronisation operation](https://github.com/elhub/flex-information-system/blob/7e1d864aaad1d3af62b8af052e478cbb7993a435/db/staging/energy_supplier_balance_responsibility_staging.sql#L83-L196), where we also replace a whole timeline of a resource with valid time, but trying to preserve what already exists.

```sql
-- need this for PG<17
DELETE FROM apeu WHERE NOT EXISTS /* [apeu from adapter with same AP - EU - valid_from] */;

MERGE INTO apeu
    USING apeu_from_adapter
    ON -- [same AP - EU - valid_from]
    
    WHEN MATCHED AND -- [valid_to or end_user_id is different]
    THEN UPDATE -- [both fields]

    WHEN NOT MATCHED /* BY TARGET (new stuff only in adapter) */
    THEN INSERT -- ...
```

enzo-crance-statnett (commented):
Looks good but I agree on the `MERGE` (at least we should try) and I need an explanation on the reentrant transaction thing.

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/AccountingPointService.kt
kenneth-strand-statnett: Refactored it to a merge into

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointQuery.kt
kenneth-strand-statnett: Fixed

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointQuery.kt
kenneth-strand-statnett: I renamed it to insertIfNotExists. The CTE is not executed if it is not referenced/used, I have tested it

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointRepository.kt
```diff
@@ -42,102 +52,370 @@ interface AccountingPointRepository {
...
+                        val validTo = endUser.validTo.toSqlTimestampOrNull()
+
+                        conn.prepareStatement(MERGE_AP_END_USER).use { stmt ->
+                            stmt.setLong(1, accountingPointId)
+                            stmt.setLong(2, endUserPartyId)
+                            stmt.setTimestamp(3, validFrom)
+                            stmt.setTimestamp(4, validTo)
+                            stmt.execute()
+                        }
+                    }
```
enzo-crance-statnett: Is it possible to give a list to the `MERGE` statement instead of looping on `endUsers`?

Now you have replaced the iteration from several statements to only one, but if we can manage to give the whole list as the source of the `MERGE`, then we replace the whole loop with only one statement. That's the real goal here IMO.

Maybe we can turn the list into a temporary view? IDK.

enzo-crance-statnett (commented):
It looks almost ready 😄

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointQuery.kt
enzo-crance-statnett: Weird. I wonder if the docs are wrong or if there is something I don't understand in the phrasing.

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointRepository.kt
kenneth-strand-statnett: Yes I agree. I tried to do that initially, but the features in java.sql are pretty poor. I'll have another look at it.

enzo-crance-statnett (dismissed):

PR #722, opened 2026-04-08
test: sqlc example in kbackend (FLEX-1064)

enzo-crance-statnett:
Example of use of sqlc in kbackend. Not meant to be merged, just a basis for discussion.

➕ no manual statement manipulation/execution/extraction, we abstract all use of SQL behind a function that takes parameters and returns what the query returns

➕ named arguments, don't need to think about indexes

➖ a bit of extra boilerplate (for example models here won't be used at all since we don't select a full resource) and need to explicit cast stuff to avoid getting `Any` in the generated code

➖ here we use the `flex` schema (so the internal tables), so `flex.sql` cannot be generated as easily as what we did in the Go backend with the `api` schema


krilor:
We did a review and the kotlin plugin for sqlc is not that good. This was a worthwhile experiment but a dead end

PR #723, opened 2026-04-08
feat: run kbackend in local Docker Compose setup (FLEX-1109)

enzo-crance-statnett:
This PR runs the Kotlin backend as part of the Docker Compose setup.

We avoid writing a custom Dockerfile by reusing the Jib plugin already present in the Kotlin projects at Elhub, We just have to check in the justfile that the image is built locally before starting the Docker Compose.

### DIFF
--- a/justfile
+++ b/justfile
@@ -24,6 +24,7 @@ tbls-lint:
 
 gradle-docker-image:
     #!/usr/bin/env bash
+    set -euo pipefail
     cd kbackend
     ./gradlew jibDockerBuild --image=flex-kbackend
     cd ..
--- a/justfile
+++ b/justfile
@@ -22,15 +22,14 @@ megalinter:
 tbls-lint:
     tbls lint
 
-gradle-docker-image:
+build-kbackend:
     #!/usr/bin/env bash
     set -euo pipefail
     cd kbackend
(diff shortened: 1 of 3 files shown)


justfile
```diff
@@ -22,8 +22,15 @@ megalinter:
 tbls-lint:
     tbls lint
 
+gradle-docker-image:
```
krilor: NIT: Maybe rename this to something with "build" and "kbackend"?
"_build-kbackend" or something?

Just to give more info about what it does

justfile
```diff
@@ -42,11 +49,11 @@ unload:
     ./db/unload.sh
 
 # docker compose pull
-pull:
+pull: gradle-docker-image
```
krilor: Doesn't the "pull_policy" on the docker-compose kbackend ensure that this dependency is not there?

justfile
```diff
@@ -22,8 +22,15 @@ megalinter:
 tbls-lint:
     tbls lint
 
+gradle-docker-image:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    cd kbackend
+    ./gradlew jibDockerBuild --image=flex-kbackend
+    cd ..
```
krilor: NIT: we don't need to cd back, I think

krilor (dismissed):

PR #733, opened 2026-04-10
refactor: set up reverse proxy from backend to kbackend for CU lookup (FLEX-1062)

enzo-crance-statnett:
This PR changes CU lookup in the Go backend from a local implementation to reverse-proxying for the kbackend implementation.

Some dead code is therefore being deleted, and scope checking is now implemented in kbackend.

We make sure the pre-existing Python API tests pass on the new implementation (WIP).


kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointRepository.kt
```diff
@@ -131,7 +132,7 @@ class AccountingPointRepositoryImpl : AccountingPointRepository {
                     id?.right()
                         ?: run {
                             logger.info { "No match for end user $endUserBusinessId and accounting point $accountingPointBusinessId" }
```
krilor, line 134: Not part of this PR but this logline with "$endUserBusinessId" must be removed. We do not want SSN in logger.info

kbackend/src/main/kotlin/no/elhub/flex/auth/Scope.kt
krilor, line 1: Shouldn't this scope stuff should have some test coverage?

krilor (dismissed):
This is looking good :) But we should add some testing of the scope stuff.

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/AccountingPointService.kt
```diff
@@ -71,23 +73,24 @@ class AccountingPointServiceImpl(
...
 
     context(principal: FlexPrincipal)
     override suspend fun checkEndUserMatchesAccountingPoint(
         endUserBusinessId: String,
         accountingPointBusinessId: String
     ): Either<AppError, Int> = accountingPointRepository.checkEndUserMatchesAccountingPoint(endUserBusinessId, accountingPointBusinessId).mapLeft { error ->
         when (error) {
-            is NotFoundError -> EndUserError("Accounting point not found")
+            is NotFoundError -> ResourceNotFoundError("Accounting point not found")
+            is NoMatchError -> EndUserError("End user does not match")
```
kenneth-strand-statnett, line 93: Is this information we want to expose?

enzo-crance-statnett: The caller probably has other ways to know whether their accounting point exists or not anyway. I think it is a better user experience to figure out more quickly that there is a typo in your accounting point ID for instance, instead of getting a 403 that actually isn't really a 403.

PR #734, opened 2026-04-10
feat: feature switch for AP sync in CU lookup (FLEX-1066)

enzo-crance-statnett:


kbackend/src/main/kotlin/no/elhub/flex/routes/controllableunit/ControllableUnitLookup.kt
```diff
@@ -54,7 +58,9 @@ class ControllableUnitLookup(
                 val validFrom = controllableUnits.minByOrNull { it.startDate }?.startDate?.asLocalStartOfDayInstant()
                     ?: Instant.atLocalStartOfToday()
 
-                accountingPointService.synchronizeAccountingPoint(accountingPointBusinessId, validFrom).bind()
+                if (accountingPointAdapterSyncEnabled) {
```
kenneth-strand-statnett, line 62: Maybe add a info or debug log statement with something like "Synchronization of accounting point is disabled"?

kenneth-strand-statnett (dismissed):

kbackend/src/main/kotlin/no/elhub/flex/routes/controllableunit/ControllableUnitLookup.kt
```diff
@@ -34,7 +35,10 @@ private val CONTROLLABLE_UNIT_BUSINESS_ID_REGEX =
 class ControllableUnitLookup(
     private val repo: ControllableUnitRepository,
     private val accountingPointService: AccountingPointService,
+    @Property("accounting-point-adapter.sync-enabled") private val accountingPointAdapterSyncEnabledStr: String = "true",
```
kenneth-strand-statnett: This is fine, but if we want to avoid the manual step of converting properties from strings everywhere we can do something like this:

```
// Properties injected with @Property
    val propertiesForKoin = listOf(
        "accounting-point-adapter.base-url" to String::class,
        "accounting-point-adapter.api-key" to String::class,
        "accounting-point-adapter.sync-enabled" to Boolean::class,
    )
    fun getPropertyValue(pair: Pair<String, KClass<*>>): Any = when (pair.second) {
        String::class -> environment.config.property(pair.first).getString()
        Boolean::class -> environment.config.property(pair.first).getString().toBoolean()
        else -> throw IllegalArgumentException("Unsupported property type: ${pair.second}")
    }

    startKoin<FlexApp> {
        environmentProperties()
        properties(propertiesForKoin.associate { it.first to getPropertyValue(it) })
        slf4jLogger()
    }
```

and then

`@Property("accounting-point-adapter.sync-enabled") private val accountingPointAdapterSyncEnabled: Boolean = true`

What do you think?

enzo-crance-statnett: I agree, it's better with typed properties and not strings everywhere. I just found another implementation as I don't fancy mentioning class types.

kenneth-strand-statnett (approved):
Looks good 👍

PR #741, opened 2026-04-14
feat: split SPGPA maximum active power into 2 fields (up/down) (FLEX-1111)

enzo-crance-statnett:
No constraints on the values for now (it will depend on the MAP at SPG level, probably a part of https://elhub.atlassian.net/browse/FLEX-1072 instead)


openapi/resources.yml
```diff
@@ -2909,18 +2909,34 @@ resources:
           en: Status
           nb: Status
           nn: Status
-      maximum_active_power:
-        description: The maximum active power applied for. Stored in kilowatts.
+      maximum_active_power_up:
+        description: >-
+          The maximum active power applied for in the upward direction. Stored
```
krilor: Not sure if possible but we could try to make this clearer on what "up" means. I think we at least should include the word "regulation direction".

> The maximum active power applied for in regulation direction up.

Its important since regulation direction is "backwards" for consumption. Meaning up regulation direction for a consumption device means using _more_ power. Its related to the frequency, not energy flow.

krilor (dismissed):

PR #747, opened 2026-04-15
fix(test): make tests great again w/ kbackend reverse-proxying

enzo-crance-statnett:
- test container for adapter copies test dataset so sync does no visible change and the rest of tests continues to work
- fix a merge commit issue that removed the HTTP 403 / HTTP 404 changes on lookup
- fix a party test where several end user parties were being created on the same entity
- some cleaning in the logs (less end user IDs displayed)

### DIFF
--- a/kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/AccountingPointService.kt
+++ b/kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/AccountingPointService.kt
@@ -91,7 +91,7 @@ class AccountingPointServiceImpl(
                 }
             },
             ifRight = { adapterAccountingPoint ->
-                logger.debug { "Raw data from adapter: $adapterAccountingPoint" }
+                logger.debug { "Raw data from adapter: ${adapterAccountingPoint.hidingEndUserIDs()}" }
                 with(FlexPrincipal.internalData()) {
                     flexTransaction { _ ->
                         either {
@@ -123,6 +123,9 @@ class AccountingPointServiceImpl(
         logger.error { "$context failed: $this" }
         return InternalServerError(traceIdOrUnknown())
     }
+
+    private fun AdapterAccountingPoint.hidingEndUserIDs(): AdapterAccountingPoint =
+        copy(endUser = endUser.map { it.copy(businessId = "<HIDDEN>") })
 
     private fun AdapterAccountingPoint.toAccountingPointEndUsers(accountingPointId: Long): List<AccountingPointEndUser> =
         endUser.map { eu ->
--- a/kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/AccountingPointService.kt
+++ b/kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/AccountingPointService.kt
@@ -83,15 +83,14 @@ class AccountingPointServiceImpl(
         .fold(
             ifLeft = { e ->
                 if (e.code == HttpStatusCode.NotFound) {
-                    logger.warn { "Accounting point not found in adapter: $accountingPointBusinessId" }
                     e.left()
                 } else {
-                    logger.warn { "Failed to fetch accounting point data for $accountingPointBusinessId — skipping sync" }
+                    logger.warn { "Skipping sync" }
                     Unit.right()
                 }
             },
             ifRight = { adapterAccountingPoint ->
-                logger.debug { "Raw data from adapter: ${adapterAccountingPoint.hidingEndUserIDs()}" }
+                logger.debug { "Raw data from adapter: ${adapterAccountingPoint.anonymizedForLogging()}" }
                 with(FlexPrincipal.internalData()) {
                     flexTransaction { _ ->
                         either {
@@ -124,8 +123,7 @@ class AccountingPointServiceImpl(
         return InternalServerError(traceIdOrUnknown())
     }
 
-    // useful for logging the structure without showing sensitive data
-    private fun AdapterAccountingPoint.hidingEndUserIDs(): AdapterAccountingPoint =
+    private fun AdapterAccountingPoint.anonymizedForLogging(): AdapterAccountingPoint =
         copy(endUser = endUser.map { it.copy(businessId = "<HIDDEN>") })
 
     private fun AdapterAccountingPoint.toAccountingPointEndUsers(accountingPointId: Long): List<AccountingPointEndUser> =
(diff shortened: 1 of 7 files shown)


local/accounting-point/main.py
```diff
@@ -21,10 +19,10 @@
...
 # Also https://elhub.github.io/flex-information-system/guides/test-data/
 
-TEST_DATA_SUFFIX_TO_ES_MGA = {
-    "000": ("1337000000020", "42Y-COMMON-SHA-W"),
-    "001": ("1337000100027", "42Y-TEST-SUITE-6"),
-    "002": ("1337000200024", "42Y-EMIL-POST--Q"),
+TEST_DATA_SUFFIX_TO_EU_ES_MGA = {
+    "000": ("13370000000", "1337000000020", "42Y-COMMON-SHA-W"),
+    "001": ("13370000001", "1337000100027", "42Y-TEST-SUITE-6"),
+    "002": ("emil.post@example.com", "1337000200024", "42Y-EMIL-POST--Q"),
```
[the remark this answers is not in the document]
enzo-crance-statnett: Valid comment, but we use email in test data so this code is not wrong. It's the kbackend that doesn't have email handling (`AccountingPointRepository.kt/resolveOrCreateEntity`).

We could add email handling or delete this case from test data (which does not break the tests -- it just means we should make sure not to lookup these accounting points in tests in the future, otherwise we'll get 404).

We probably have PID/ORG assumed a bit all over the place, so I vote for 2 and comment this line for now.

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/AccountingPointService.kt
```diff
@@ -80,11 +81,17 @@ class AccountingPointServiceImpl(
         validFrom: Instant
     ): Either<AppError, Unit> = fetchAccountingPointData(accountingPointBusinessId, validFrom)
         .fold(
-            ifLeft = {
-                logger.warn { "Failed to fetch accounting point data for $accountingPointBusinessId — skipping sync" }
-                Unit.right()
+            ifLeft = { e ->
```
kenneth-strand-statnett, line 84: This is already covered in `fetchAccountingPointData`

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/AccountingPointService.kt
```diff
@@ -117,6 +124,10 @@ class AccountingPointServiceImpl(
         return InternalServerError(traceIdOrUnknown())
     }
 
+    // useful for logging the structure without showing sensitive data
+    private fun AdapterAccountingPoint.hidingEndUserIDs(): AdapterAccountingPoint =
```
kenneth-strand-statnett: Maybe a generic `anonymized()` better describes the purpose and can be used for other entities as well?

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointRepository.kt
```diff
@@ -207,8 +208,8 @@ class AccountingPointRepositoryImpl : AccountingPointRepository {
                 DatabaseError("Failed to check end user")
             }.flatMap { id ->
                 id?.right() ?: run {
-                    logger.info { "No match for end user $endUserBusinessId and accounting point $accountingPointBusinessId" }
-                    NotFoundError("End user does not match accounting point / controllable unit").left()
+                    logger.info { "No match between end user and accounting point $accountingPointBusinessId" }
+                    NoMatchError("End user does not match accounting point / controllable unit").left()
```
kenneth-strand-statnett, line 212: What is the difference between _No Match_ and _Not Found_?

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/AccountingPointService.kt
```diff
@@ -143,7 +154,8 @@ class AccountingPointServiceImpl(
         accountingPointBusinessId: String
     ): Either<AppError, Long> = accountingPointRepository.checkEndUserMatchesAccountingPoint(endUserBusinessId, accountingPointBusinessId).mapLeft { error ->
         when (error) {
-            is NotFoundError -> EndUserError("Accounting point not found")
+            is NotFoundError -> ResourceNotFoundError("Accounting point not found")
```
kenneth-strand-statnett: Will `NotFoundError` ever be returned?

kenneth-strand-statnett (changes requested):
Looks good, just a few questions I'd like you to have a look at before approval

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointRepository.kt
enzo-crance-statnett: The main semantic reason I see is that if something is not matching, it means it _was_ found and then we ran a check on it that returned false. So returning _Not Found_ on no match would be weird. We need another error type to cover the no match case.

Regarding the error codes, _Not Found_ is naturally a 404, but _No Match_ is a 403: here having a match between AP ID and end user ID is a way to authorise the lookup operation. The user shows they know enough, otherwise they _do not have the right_ to get a lookup answer, so it is definitely a _Forbidden_ type of response.

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/AccountingPointService.kt
enzo-crance-statnett: It was _already_ never being returned before. It is a `RepositoryError` and the end errors in the handlers are the higher-level `AppError`.

`ResourceNotFoundError` is just an equivalent in the more general error type.

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/AccountingPointService.kt
enzo-crance-statnett: The log is indeed duplicated, I fixed that.

But the `if` block here is important. If the `ifLeft` block just returns `Unit.right()`, we silence all errors (including 404) and skip the sync. Then we run the end user check which always fails with the same error no matter whether the AP exists or not. Without a special case propagating the 404, we cannot identify the not found case IIUC, which was the point of this addition.

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/AccountingPointService.kt
enzo-crance-statnett: Now I understand. The underlying function no longer gives `NotFoundError`... Thanks!

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointRepository.kt
kenneth-strand-statnett: Alright. I find it a bit weird to introduce a variant of not found. The error type from the repository is not related to the http code we return. In my mind Not Found = No rows matched the search criteria.

This is just nitpicking though :sweat_smile:  Remember to update error type in the interface doc.

kenneth-strand-statnett (dismissed):

PR #752, opened 2026-04-16
feat:  redesign comments (FLEX-1123)

s0lvang:


frontend/src/components/CommentFeed.tsx
```diff
@@ -0,0 +1,159 @@
...
+              onValueChange={(v) =>
+                setVisibility(
+                  v as ServiceProvidingGroupProductApplicationCommentVisibility,
+                )
+              }
+              placeholder="Visibility"
+            >
+              <SelectContent>
+                <SelectItem value="same_party">
+                  Internal (same party)
```
krilor: These texts should probably use translated enum keys?

frontend/src/components/CommentFeed.tsx
```diff
@@ -0,0 +1,159 @@
...
+            <p className="text-sm text-red-600 mt-1">
+              Failed to post comment. Please try again.
+            </p>
+          )}
+          <div className="flex justify-end items-center gap-2 mt-2">
+            <Select
+              value={visibility}
+              onValueChange={(v) =>
+                setVisibility(
+                  v as ServiceProvidingGroupProductApplicationCommentVisibility,
```
krilor: Using ServiceProvidingGroupProductApplicationCommentVisibility in a general component feels a bit off. We should consider making the enum shared (aka move it to resource-base.yml) across the comments resources.

krilor: I realize now that the problem is "bigger" than just here. SPGPA present in both CommentFeed and CommentBubble. Is there a way to make it truly reusable?

krilor (dismissed):
Good stuff. Just some notes about reusability that we should consider adressing.

frontend/src/components/CommentFeed.tsx
s0lvang: We can create a type that is not autogenerated which all commentTypes satisfies. I just see it as unlikely that the different comment types will deviate. In the current implementation, if you send in an SPPA-comment it will not fail since the types are equal

krilor: Ok.  The different comment types generally differ in one fk field (what parent resource it is pointing to). It has different names on each of the things.

We should maybe add this behaviour that as a whycomment?

I'm not sure I want to go down that road again but we could also have created inheritance in the openapi spec (and thus get generated types) as well with a common base and just add the additional field for each of the specific comment resources.

krilor (approved):

PR #754, opened 2026-04-17
docs: Added resource documentation for Accounting Point Grid Location (FLEX-1136)

kenneth-strand-statnett:


docs/resources/accounting_point_grid_location.md
```diff
@@ -0,0 +1,135 @@
...
+
+The electrical (topological) location of an accounting point in the common grid
+model ([Nemo](https://nemo.elbits.no/modell/)). Since a controllable unit is
+always behind an accounting point, this also gives the grid location for all
+controllable units connected to that accounting point.
+
+The connecting system operator (CSO) is responsible for registering and
+maintaining the grid location. Another system operator may also register a grid
+location, but only when the current entry is missing or has been set by the
+system or guessed by another system operator. The system will automatically
```
krilor: This part is will not be implemented in the current story/work and we should probably just keep that in the design for now?

docs/resources/accounting_point_grid_location.md
```diff
@@ -0,0 +1,135 @@
...
+| <a name="field-accounting_point_id" href="#field-accounting_point_id">accounting_point_id</a>          | The accounting point this grid location belongs to.                          | bigint<br/>Required<br/>Non-updatable                                                     | [accounting_point.id](accounting_point.md#field-id) |
+| <a name="field-object_type" href="#field-object_type">object_type</a>                                  | The type of object in the common grid model that the accounting point is at. | text<br/>One of: `substation`, `transformer`<br/>Required                                 |                                                     |
+| <a name="field-business_id" href="#field-business_id">business_id</a>                                  | Business identifier (mRID) referencing the object in the common grid model.  | text<br/>Pattern: `^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$` |                                                     |
```
krilor: This pattern is a bit too strict. I don't think we should require UUID v4. I know for a fact that UUID v3 at least is also in use.

docs/resources/accounting_point_grid_location.md
```diff
@@ -0,0 +1,135 @@
...
+| <a name="field-name" href="#field-name">name</a>                                                       | Name of the grid model object at the location.                               | text<br/>Required<br/>Max length: `512`                                                   |                                                     |
+| <a name="field-nominal_voltage" href="#field-nominal_voltage">nominal_voltage</a>                      | Nominal voltage level at the grid location, in kilovolt (kV).                | decimal<br/>Required<br/>Min: `0`                                                         |                                                     |
```
krilor: We are trying to make the description on where "unit fields" work better with unit input where units can be changed. The description is used in UI/API docs etc. What do we do in practice?

We write "Stored in kilovolt (kV)."

So 

> Nominal voltage level at the grid location. Stored in kilovolt (kV).

docs/resources/accounting_point_grid_location.md
```diff
@@ -0,0 +1,135 @@
...
+| <a name="field-recorded_by" href="#field-recorded_by">recorded_by</a>                                  | The identity that recorded the resource.                                     | bigint<br/>Read only                                                                      |                                                     |
+
+## Validation Rules
+
+| Validation rule key | Validation rule                                                                                                             | Status |
+|---------------------|-----------------------------------------------------------------------------------------------------------------------------|--------|
+| APGL-VAL001         | Updates to the grid location are only allowed based on the current `source` value. See transition table below.              | TODO   |
+| APGL-VAL002         | `quality=confirmed` is only permitted when `source` is `cso` or `grid_model`. All other sources must use `quality=guessed`. | TODO   |
```
krilor: `so` should also be able to use "confirmed", imho. E.g. if a PSO has talked to the CSO.

docs/resources/accounting_point_grid_location.md
```diff
@@ -0,0 +1,135 @@
...
+### APGL-VAL001 source transition table
+
+The table shows which new `source` values are permitted given the current value
+of the record. `yes` means the transition is allowed, `no` means it is denied.
+
+| Current ↓ \ New → | `grid_model` | `cso` | `so`                                          | `system` |
+|-------------------|--------------|-------|-----------------------------------------------|----------|
+| `grid_model`      | yes          | no    | no                                            | no       |
+| `cso`             | yes          | yes   | no                                            | no       |
+| `so`              | yes          | yes   | yes, but not transition `confirmed`→`guessed` | no       |
```
krilor: I'm second-guessing this "but not transition `confirmed`→`guessed`" part. 

What if SO does a mistake and sets "confirmed" while meaning "guessed". They cannot fix that.

Lets remove it and assume that parties will behave? Avoiding an unneccesary rule is nice :)

docs/resources/accounting_point_grid_location.md
```diff
@@ -0,0 +1,135 @@
...
+| `system`          | yes          | yes   | yes                                           | yes      |
+| missing           | yes          | yes   | yes                                           | yes      |
+
+## Notifications
+
+No notifications.
```
krilor: We should def add notifications. I think both CSO and PSO should get notifications.

docs/resources/accounting_point_grid_location.md
```diff
@@ -0,0 +1,135 @@
...
+
+#### Organisation
+
+No policies.
+
+#### System Operator
+
+| Policy key | Policy                                                                                      | Status |
+|------------|---------------------------------------------------------------------------------------------|--------|
+| APGL-SO001 | Read accounting point grid locations where they are SO                                      | TODO   |
```
krilor: We should be more specific here? Maybe

> Create, read and update accounting poing grid locations on AP where they are CSO, ISO or PSO.

docs/resources/accounting_point_grid_location.md
```diff
@@ -0,0 +1,135 @@
...
+| APGL-SO002 | Create and update accounting point grid locations where they are SO, subject to APGL-VAL001 | TODO   |
```
krilor: The access rules should be logically and technically separate from the validation rules. This is a strategy to make the implementation clean/simple.

One way to think about it is to consider it different "layers".

docs/resources/accounting_point_grid_location.md
```diff
@@ -0,0 +1,135 @@
...
+| APGL-SO003 | Read history on accounting point grid locations                                             | TODO   |
```
krilor: SO003 is missing qualification. What SO can read history?

docs/resources/accounting_point_grid_location.md
krilor: Are we missing validation rules for the use of source?

* SO can only use `so` and `cso`.
* For a SO to use `cso` we must validate that they in fact are CSO.
* For a SO to use `so` we must validate that they are PSO.

docs/resources/accounting_point_grid_location.md
```diff
@@ -0,0 +1,135 @@
...
+|--------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------|-----------------------------------------------------|
+| <a name="field-id" href="#field-id">id</a>                                                             | Unique surrogate key.                                                        | bigint<br/>Read only                                                                      |                                                     |
+| <a name="field-accounting_point_id" href="#field-accounting_point_id">accounting_point_id</a>          | The accounting point this grid location belongs to.                          | bigint<br/>Required<br/>Non-updatable                                                     | [accounting_point.id](accounting_point.md#field-id) |
```
krilor: We should note somehow that AP ID has a unique constraint, making this a one-to-one rel

krilor (commented):
Good stuff. See comments :)

docs/resources/accounting_point_grid_location.md
kenneth-strand-statnett: Do we need this as input? Should be possible to derive this from the request?

krilor: Yep. SO means "flex_system_operator" role

docs/resources/accounting_point_grid_location.md
kenneth-strand-statnett: Added section

docs/resources/accounting_point_grid_location.md
kenneth-strand-statnett: :+1: Fixed

docs/resources/accounting_point_grid_location.md
kenneth-strand-statnett: Fixed

docs/resources/accounting_point_grid_location.md
kenneth-strand-statnett: Fixed

docs/resources/accounting_point_grid_location.md
kenneth-strand-statnett: I removed the pattern validation as we do  not know the format and it may change in the future

docs/resources/accounting_point_grid_location.md
kenneth-strand-statnett: Updated description

docs/resources/accounting_point_grid_location.md
kenneth-strand-statnett: Removed it :+1:

docs/resources/accounting_point_grid_location.md
krilor: UUID validation is fine (and wanted). Its just that we cannot limit it to UUID4

docs/resources/accounting_point_grid_location.md
krilor: Should update the linked documentation/design as well.

docs/resources/accounting_point_grid_location.md
kenneth-strand-statnett: Fixed

docs/resources/accounting_point_grid_location.md
kenneth-strand-statnett: Do we need to update anything here? We don't need any validation rules for this if we remove source from create/update payload and instead derive it form the request.

docs/resources/accounting_point_grid_location.md
kenneth-strand-statnett: Fixed

docs/resources/accounting_point_grid_location.md
kenneth-strand-statnett: Added pattern that supports all UUID versions.

PR #763, opened 2026-04-22
feat: kotlin CRUD implementation

kenneth-strand-statnett:
Example of how AP Grid Location can be implemented in the kotlin backend


krilor:
This was a worthwhile exersice but its not the time for it. closing it for now and we can keep it as a reference going forward.

PR #764, opened 2026-04-22
feat: add controllable unit summary resource (FLEX-1143)

enzo-crance-statnett:


openapi/openapi-api-no-default.json
```diff
@@ -1465,6 +1469,85 @@
...
+                        "description": "Total number of technical resources in the controllable unit.",
+                        "format": "bigint",
+                        "type": "integer",
+                        "readOnly": true,
+                        "example": 3
+                    },
+                    "count_technical_resource_by_technology": {
+                        "description": "Number of technical resources in the controllable unit, broken down by technology. Keys are technology IDs, values are counts.",
+                        "type": "object",
+                        "additionalProperties": {
```
krilor: Maybe we can use actual enums? https://www.jsonschemavalidator.net/s/AoCTw1Np

enzo-crance-statnett: I had not seen technology was an enum. Thought there were too many cases and that was the reason for the `flex.technology` table.

openapi/resources.yml
```diff
@@ -396,6 +396,139 @@ resources:
...
+                  properties:
+                    count:
+                      type: integer
+                    maximum_active_power:
+                      $ref: "#/components/schemas/numeric_aggregation"
+              by_technology:
+                type: object
+                propertyNames:
+                  enum:
+                    - hydropower
```
krilor: I feel like it should be possible to use the existing enum (in openapi-api-base). Did we try that?
Ref https://github.com/elhub/flex-information-system/pull/764#discussion_r3123977432 ?

local/scripts/openapi_to_db.py
```diff
@@ -97,7 +97,7 @@ def collect_relationships(resources):
         child = resource["id"]
         props = resource.get("properties", {})
         for field, attr in props.items():
-            if not field.endswith("_id"):
```
krilor, line 100: Why this change?

local/scripts/templates/embedding.j2.sql
```diff
@@ -8,4 +8,4 @@ CREATE OR REPLACE FUNCTION api.{{ rel.name }}(
 RETURNS SETOF api.{{ rel.parent.resource }}{% if rel.cardinality == 'one' %} ROWS 1{% endif %} AS $$
   select * from api.{{ rel.parent.resource }} where {{ rel.parent.name }} = $1.{{ rel.child.name }}
```
krilor, line 9: There should be a relationship or two here.

I guess it is because we did not add a "controllable_unit_id" foreign key field. I think we should do that. It helps keep the API consistent. It is "weird" that the FK is on the ID field (?). The "cost" is just a duplication of the id into controllable_unit_id.

local/scripts/openapi_to_db.py
enzo-crance-statnett: Allows using `id` as a foreign key. But the consistency argument is good enough so we revert changes to the scripts and add a `controllable_unit_id` field that copies `id`.

openapi/resources.yml
```diff
@@ -390,6 +390,125 @@ resources:
...
+      en: Controllable Unit Summary
+      nb: Sammendrag av kontrollerbar enhet
+      nn: Samandrag av kontrollerbar eining
+    operations:
+      - read
+    generate_views: false
+    audit: false
+    properties:
+      id:
+        description: Unique surrogate key.
```
krilor: Should problably note that this = CU ID. So that people know that they can use /cusummary/cuid

krilor (dismissed):
Nice!

PR #776, opened 2026-04-24
feat: add service providing group summary resource (FLEX-1145)

enzo-crance-statnett:


backend/data/static/openapi.json
```diff
@@ -1816,6 +1820,163 @@
...
+                    },
+                    "service_providing_group_id": {
+                        "description": "The ID of the service providing group this resource is a summary of.",
+                        "format": "bigint",
+                        "type": "integer",
+                        "readOnly": true,
+                        "example": 4
+                    },
+                    "controllable_unit": {
+                        "description": "Aggregated statistics on controllable units currently in the service providing group, including count and maximum active power breakdowns (sum, average, min, max) by category and technology.",
```
krilor: I don't think we have "by category and technology"

enzo-crance-statnett: Indeed, it's probably a typo. We said we would not do the breakdown on the CU level.

krilor (dismissed):
Sweet!

PR #780, opened 2026-04-28
feat: tweaks after user test (FLEX-1157)

s0lvang:
- removed accordion from cu input
- Dont use abbreviations in navigation
- removed_qualified at from sppa input
- Larger headers
- Alert when you dont have a product application on spg


frontend/src/components/Header/HeaderNav.tsx
```diff
@@ -3,8 +3,8 @@ import { Nav } from "../ui";
 
 const navLinks = [
   { label: "Dashboard", to: "/" },
-  { label: "CUs", to: "/controllable_unit" },
-  { label: "SPGs", to: "/service_providing_group" },
+  { label: "Controllable Units", to: "/controllable_unit" },
```
krilor: NIT: Should we stick to capitalisation or not? Not mix

frontend/src/service_providing_group/product_application/ServiceProvidingGroupProductApplicationInput.tsx
```diff
@@ -128,12 +127,6 @@ export const ServiceProvidingGroupProductApplicationInput = () => {
           description
           tooltip={false}
         />
-        <DateTimeInput
-          {...fields.prequalified_at}
```
krilor, line 132: Why are we removing this? Updating these timestamps should be possible. Maybe we should set props to show the now and clear buttons or someting.

frontend/src/service_providing_group/show/ServiceProvidingGroupAlerts.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+const useServiceProvidingGroupAlerts = (
+  spg: ServiceProvidingGroup,
+): AlertType | null => {
+  const { data: productApplications } = useSpgProductApplications(spg.id);
+
+  if (spg.status === "new") {
+    return {
+      severity: "info",
+      content: "The service providing group is not active.",
```
krilor, line 28: Should these alerts be party specific? I'm thinking these only make sense for the SP on the SPG.

frontend/src/service_providing_group/show/ServiceProvidingGroupAlerts.tsx
```diff
@@ -0,0 +1,49 @@
+import { ReactNode } from "react";
+import { ServiceProvidingGroup } from "../../generated-client";
+import { Alert, BodyText } from "../../components/ui";
+import { useSpgProductApplications } from "./useSpgProductApplications";
+
+type AlertType = {
```
krilor, line 7: Should we maybe try to make alerts use this pattern across the board?

krilor (commented):
Generally looks good but I am concerned about the removal of those datetime fields.

frontend/src/service_providing_group/product_application/ServiceProvidingGroupProductApplicationInput.tsx
s0lvang: I can remove it based on access. But it does not make sense to me that this timestamp should be set by the user, it should be a result of the action that sets the state to prequalifed.

frontend/src/service_providing_group/show/ServiceProvidingGroupAlerts.tsx
s0lvang: Good point

frontend/src/service_providing_group/product_application/ServiceProvidingGroupProductApplicationInput.tsx
s0lvang: I remove all fields that you dont have access to now

enzo-crance-statnett (dismissed):

PR #782, opened 2026-04-28
feat(deploy): set a specific liquibase context for loading test_data schema (FLEX-1160)

enzo-crance-statnett:
Activated on local.

`@` means that no further setup in Helm = no `test_data` schema in the environment.

### DIFF
--- a/db/changelog.yml
+++ b/db/changelog.yml
@@ -51,7 +51,7 @@ databaseChangeLog:
       file: ./test_data/test_data.sql
       relativeToChangelogFile: true
       # `@` means run only when the context is explicitly set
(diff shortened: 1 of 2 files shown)


krilor (dismissed):

db/changelog.yml
```diff
@@ -50,6 +50,8 @@ databaseChangeLog:
   - include:
       file: ./test_data/test_data.sql
       relativeToChangelogFile: true
+      # `@` means run only when the context is explicitly set
```
krilor, line 53: Should we consider using a more general context like "test" ?

PR #788, opened 2026-04-29
fix(ui): make more space for SPG aggregates (FLEX-1072)

enzo-crance-statnett:
Showing both the table and the diagram on the same line was a bit cluttered. Switching to tabs should fix the diagram label issue and make the table more readable at the same time.

Technology:

Screenshot 2026-04-29 164009

Category:

Screenshot 2026-04-29 164017

An alternative is to always show both, but one on top of the other.


krilor (approved):
🥳

PR #795, opened 2026-04-30
feat(backend): scope check on resource embedding (FLEX-1161)

enzo-crance-statnett:
This PR adds a scope check on the resource embedding query parameter, so that the user cannot embed resources their scopes do not cover.

The solution involves splitting the existing methods for turning `embed` into PostgREST `select`, into 2 phases (parsing `embed`, and writing `select`), which lets us use the parsed structure before writing it back to the query parameters. We traverse it and get all the resources mentioned, then we check the scopes in the token cover all of them, otherwise the user gets a 403.

Unit testing was done on the new methods. API testing is very primitive for now, because the API client only allows a small set of scopes. We should at some point extend the support for scopes to all the possible combinations. This is out of scope because it implies a bit more work (cf https://elhub.atlassian.net/browse/FLEX-1175).


test/auth_tests/test_scopes.py
```diff
@@ -413,3 +414,95 @@ def test_scopes_jwt_bearer_membership_party(data):
...
+    # embed resources in a query
+
+    sts = SecurityTokenService()
+
+    client_ent = cast(AuthenticatedClient, sts.get_client(TestEntity.TEST))
+    client_fiso = cast(AuthenticatedClient, sts.get_client(TestEntity.TEST, "FISO"))
+
+    ent_id = sts.get_userinfo(client_ent)["entity_id"]
+
+    with open("./test/keys/.test.pub.pem") as f:
```
krilor: Should we maybe put jwt stuff somewhere common and not in the test?

backend/data/data.go
```diff
@@ -608,6 +614,37 @@ func (data *api) postgRESTHandler(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	if len(embedNodes) > 0 { // embed parameter exists -> check scopes
```
krilor: NIT: We could consider moving the actual scope check to a separate function to avoid cyclop and separate concerns between the modules.

krilor (changes requested):
This change seems to assume that embed key = resource name. It is sadly not  that simple :(

Examples:

* On CU, CUSP is embedded as "service_provider" (and not "controllable_unit_service_provider")
* On CUSP, the SP party is imbedded as "service_provider" (and not "party")

I think we need to generate some refdata/code to know the resource from the embed key.
An alternative is to generate models from the openapi spec and use object names of the embed keys to get the resource.

enzo-crance-statnett:
Generated a structure from the resource YAML. Reusing the relationships package in python

krilor (approved):

PR #802, opened 2026-05-05
feat(db): status field on metering grid area (FLEX-1171)

enzo-crance-statnett:
This PR adds a `status` field on MGA with two possible values, `active` or `inactive`, allowing soft deletes when syncing data from external sources.

Opening draft for discussion because I am unsure of the behaviour we should aim for in the various places in the system. Just trying to identify the potential impact of such an update and list the tables/views/functions we need to think about.


enzo-crance-statnett:
Filtering on active MGA in lookup AP sync means we probably get an error if the sync is tried with an inactive MGA. Haven't checked if this can happen and what it implies, I need to see more in the details.

Filtering on active MGAs in the views will introduce quite a hard filter where, as we get data going through the MGA, if an AP belongs to an MGA marked inactive after a MGA sync, then users basically lose access to everything about this AP (AP-BRP, AP-BZ, AP-SO, etc, will filter out all the rows).

IDK if it may break stuff in the frontend: any technical MGA ID left somewhere won't be translatable to a EIC-Y code for instance if the MGA is inactive, so we need to make sure the filters are everywhere and there is no "escape". But if we filter everywhere else it may be a consistent choice. Basically an inactive MGA "disappears" from the system, which was kind of the point of a soft delete.

On ESBR staging I just changed the view resolving the foreign keys, so that basically any data coming from the external source about dead MGAs is ignored. Probably expected as well, just making it clear.

enzo-crance-statnett:
Switched to an approach where we just open `status` in the API but we don't use it for anything else, it is basically just information for now.

No frontend change because an inactive MGA will probably lead to changes in the next AP sync which means that the current MGA is almost always an active one. At least it is most likely out of scope for now.

krilor (approved):

PR #803, opened 2026-05-05
feat: Implemented validation rules for source/quality. Rules does not apply to FISO (FLEX-1180)

kenneth-strand-statnett:


enzo-crance-statnett (changes requested):
Looks good, requesting changes for API tests to validate the behaviour

db/flex/accounting_point_grid_location.sql
```diff
@@ -87,6 +87,61 @@ ON accounting_point_grid_location
...
+        ) THEN
+            RAISE EXCEPTION
+                'Invalid source transition from % to %.', OLD.source, NEW.source;
+        END IF;
+    END IF;
+    RETURN NEW;
+END;
+$$;
+
+CREATE OR REPLACE TRIGGER b_accounting_point_grid_location_source_transition_check
```
krilor, line 118: Shouldn't trigger create be its own changeset?

kenneth-strand-statnett: It is, or am I misunderstanding something?

enzo-crance-statnett (dismissed):
Not so easy to test as source is set by the system. Agreed to merge as is.

db/flex/accounting_point_grid_location.sql
krilor: Not the way I ment it at least 😅

By changeset I mean that it has its own `-- changeset` header. One for the function, another for the trigger.

Like https://github.com/elhub/flex-information-system/blob/main/db/flex/controllable_unit.sql#L72-L101

Just a nitpick but it is generally encuraged to have small changesets in liquibase to aid in debugging errors etc.

kenneth-strand-statnett: Ah, separate for trigger and function - gotcha.
Fixed

PR #806, opened 2026-05-05
ci: manual merge TeamCity pipeline for prod1 (like euro) (FLEX-1115)

enzo-crance-statnett:


.teamcity/settings.kts
```diff
@@ -73,6 +73,22 @@ elhubProject(Group.FLEX, "flex-information-system") {
...
+                    }.triggerOnVcsChange {
+                        triggerRules = """
+                                -:*
+                                +:backend/**
+                                +:db/**
+                        """.trimIndent()
+                    }
+                    gitOps {
+                        buildNameSuffix = "backend prod1"
+                        clusters = setOf(KubeCluster.PROD1)
```
krilor: Instead of adding an additonal manual step here, maybe we should add KubeCluster.PROD1 to the existing "backend euro" ones? Meaning that we do:

* flex-test ( test9 ) - automerge
* flex-euro (mt1) and flex-prod1 (prod1) - manual merge but in a single PR?

Not sure if it makes sense but I just want to avoid the toil of manual approval

krilor (approved):

PR #807, opened 2026-05-05
feat: party staging w/ unlogged table (FLEX-1116)

krilor:


enzo-crance-statnett (approved):
LGTM.

There is some confusion with `party` in the fact that we have an extra landing table in the `flex` schema (so there are 2 hops from external data to `flex.party`), and that it was unfortunately called `party_staging` where we now have a `staging` schema and now a `staging.party` table...

I guess the `flex_` prefix is a way to make things a bit clearer, and I do not think this PR is responsible for the confusion, but it is just a quick thought I needed to share 😅

krilor:
> LGTM.
> 
> There is some confusion with `party` in the fact that we have an extra landing table in the `flex` schema (so there are 2 hops from external data to `flex.party`), and that it was unfortunately called `party_staging` where we now have a `staging` schema and now a `staging.party` table...
> 
> I guess the `flex_` prefix is a way to make things a bit clearer, and I do not think this PR is responsible for the confusion, but it is just a quick thought I needed to share 😅

Indeed. That is also why I tried updating the drawing/docs to show whats going on.

PR #813, opened 2026-05-07
feat: more restrictive RLS policies on accounting point (FLEX-1167)

krilor:


enzo-crance-statnett (dismissed):

krilor:
Approved (not allowed to press the button since I was the one that created the PR 😅)

PR #815, opened 2026-05-07
fix: adjust notice to the non unique party.business_id (FLEX-1182)

enzo-crance-statnett:


db/flex/notice/notice_party.sql
```diff
@@ -31,7 +31,7 @@ WITH (security_invoker = false) AS (
                 )
             )
         ) AS data, -- noqa
-        md5(p_stg.gln::text) AS deduplication_key -- noqa
+        md5(p_stg.gln::text || p_stg.type) AS deduplication_key -- noqa
```
enzo-crance-statnett, line 34: > [!NOTE]
> This is needed because two parties can be missing with the same GLN. Updating the hash means existing cases like this one will be set to resolved and new notices will be opened. Probably an edge case anyway...

krilor: Nice! Seems like the manual-deduplication-key-per-notice-type was a good design choice in the end 👍

krilor (approved):
LGTM

PR #819, opened 2026-05-08
feat: dashboard v1.5

s0lvang:
image


frontend/src/dashboard/hooks/useControllableUnits.ts
```diff
@@ -0,0 +1,26 @@
...
+    enabled: partyID != null,
+    queryFn: () =>
+      listControllableUnitServiceProvider({
+        query: {
+          service_provider_id: `eq.${partyID}`,
+          valid_at: formatISO(new Date()),
+          embed: "controllable_unit",
+        },
+        headers: {
+          Prefer: "count=exact",
```
enzo-crance-statnett, line 22: I don't think we document this in the OpenAPI spec. Once more a PostgREST feature bubbling through because we don't empty the headers. 😄

It is useful but I think we should at least put a link to the [PostgREST docs](https://docs.postgrest.org/en/latest/references/api/pagination_count.html#exact-count) or IDK.

enzo-crance-statnett (approved):
This needs a functional review as well but LGTM

PR #827, opened 2026-05-11
feat: Added discriminator property for notices (FLEX-1101)

kenneth-strand-statnett:


backend/data/static/openapi.json
```diff
@@ -528,6 +542,13 @@
             "notice_data_party_outdated": {
                 "description": "Format of the data field in a notice of type no.elhub.flex.party.outdated",
                 "properties": {
+                    "notice_type": {
```
krilor: Should we use `const` in OpenAPI 3.1? It has a much clearer intent than the use of a single enum IMHO.

backend/data/static/openapi.json
```diff
@@ -566,7 +594,16 @@
                     {
                         "$ref": "#/components/schemas/notice_data_product_type_not_qualified"
                     }
-                ]
+                ],
+                "discriminator": {
+                    "propertyName": "notice_type",
+                    "mapping": {
+                        "no.elhub.flex.controllable_unit_service_provider.valid_time.outside_contract": "#/components/schemas/notice_data_valid_time_outside_contract",
```
krilor: There should be a clear one-to-one mapping between the key and the schemas IMHO. I think we should improve naming a bit here to avoid future conflicts. What happens if we have another "something.something.valid_time.outside_contract". We would have to break the naming somehow - either break clients or take on some kind of inconsistency in naming.

In my head we should avoid using the notice.type as a duplicated discriminator, but rather use a custom "type" for this. This forces us to be explicit in the individual notice views - which is a good thing in the long run imho.

There is also a bit of room for consolidation at the moment. `notice_data_party_missing`, `notice_data_party_outdated` basically share the same schema.

In total, I'd suggest something like

```json5
"discriminator": {
                    "propertyName": "type",
                    "mapping": {
                        "notice.data.valid_time.outside_contract": "#/components/schemas/notice_data_valid_time_outside_contract",
                        "notice.data.party": "#/components/schemas/notice_data_party",
                        "notice.data.product_type.not_qualified": "#/components/schemas/notice_data_product_type_not_qualified"
```

The alternative would be to actually name schemas per notice type.
So mappings would look like.

```
 "no.elhub.flex.controllable_unit_service_provider.valid_time.outside_contract": "#/components/schemas/notice_data_controllable_unit_service_provider_valid_time_outside_contract",
```

I prefer the former, but both are ok.

openapi/openapi-api-no-default.json
```diff
@@ -566,7 +594,16 @@
...
+                "discriminator": {
+                    "propertyName": "notice_type",
+                    "mapping": {
+                        "no.elhub.flex.controllable_unit_service_provider.valid_time.outside_contract": "#/components/schemas/notice_data_valid_time_outside_contract",
+                        "no.elhub.flex.party.missing": "#/components/schemas/notice_data_party_missing",
+                        "no.elhub.flex.party.outdated": "#/components/schemas/notice_data_party_outdated",
+                        "no.elhub.flex.service_provider_product_suspension.product_type.not_qualified": "#/components/schemas/notice_data_product_type_not_qualified"
+                    }
+                }
             },
```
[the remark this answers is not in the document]
krilor: We should do this 👍

krilor (approved):
🦄

PR #829, opened 2026-05-12
docs: first pass of queryparams and embed docs (FLEX-1177)

s0lvang:
short and sweet.


docs/api/query-parameters.md
```diff
@@ -0,0 +1,112 @@
...
+GET /controllable_unit?embed=accounting_point
+GET /controllable_unit?embed=accounting_point!
+GET /controllable_unit?embed=accounting_point!(bidding_zone,system_operator)
+GET /controllable_unit?embed=accounting_point!(bidding_zone!(metering_grid_area))
+GET /controllable_unit?embed=accounting_point,technical_resource
+```
+
+### Available relations
+
+Available relations are defined per parent resource in [`embed_relations_gen.go`](../../backend/data/embed_relations_gen.go).
```
krilor: We maybe should not have this in external-facing user docs?

docs/technical/query-parameters.md
```diff
@@ -0,0 +1,112 @@
+# Query Parameters
+
+The API is backed by [PostgREST](https://postgrest.org/). Most query parameters
+are forwarded to PostgREST directly; a few are rewritten by the Go proxy before
```
krilor: I think we should avoid phrasing like "the Go proxy" in userfacing docs?

docs/api/query-parameters.md
s0lvang: It was written as user-facing, but put it into the wrong location :D

krilor (commented):
1. The doc has a few implementation details not relevant for users of the API. Should we remove that and focus this on API users?
2. Should maybe make a new section in our docs called API, where we can put stuff like this (+ maybe API design) Its kind of hidden within "technical". Ref https://elhub.github.io/flex-information-system/technical/ (we probably should lift the "Auth" out from "Technical" as well)

docs/technical/query-parameters.md
s0lvang: Yeah i agree.

But i want to make clear that they are rewritten so they can rely on the postgrest docs for embed

krilor (dismissed):
I was probably not as precise as I should have been in my previous comment, but I think that "Auth" should live side-by-side with "API". So menu would be

* API
    * API design
    * Query parameters
    * FAQ
* Auth
    * Methods
    * ..
    * Model

Auth is kindof common informasjon for both api and portal users.

How we can move it prior to merge.

krilor (approved):

PR #839, opened 2026-05-15
feat: delay SPG-GP creation until SPG-PA is set to the right status (FLEX-1088)

enzo-crance-statnett:
Changes:

- new set of statuses for **SPG product application**: the SO chooses between `prequalification` and `temporary_qualified` to decide if they want to go for prequalification or validation
- trigger previously on SPG activation to cause creation of **SPG grid prequalification** resources moved to when an SPG-PA moves away from `requested`
- the trigger now launches a general sync process on SPG-GP, checking every SPG: there must be an SPG-PA that is no longer `requested` nor `rejected`, and every CSO on the CUs should have a SPG-GP -- future checks to come later
- new policy so that SO can read SPG-PA when they are targeted by a SPG-GP, now that the dependency "turned around"
- flow in SPG-PA page: when `requested`, be proposed to mark `prequalification`; when in `prequalification`, be proposed to mark `prequalified`
- warn SO that they need to do something within 4 weeks on SPG-PA creation, in the show page
- updated tests and cleanup of dead code


enzo-crance-statnett:
Screenshot 2026-05-19 123504

justfile
```diff
@@ -63,6 +63,9 @@ load: liquibase
 
     psql -X -v ON_ERROR_STOP=1 -d flex -U postgres -f ./local/postgres/users.sql
 
+    psql -X -v ON_ERROR_STOP=1 -d flex -U postgres \
+        -c "ALTER ROLE postgres SET search_path TO flex, public;"
```
enzo-crance-statnett, line 67: Got errors on `just connect`, probably PG17 default config that is a bit stricter.

I'm just modifying the `postgres` role here and not the whole DB so it doesn't apply to all the DB connections we do in the backend etc. Otherwise we may hide some errors by using `ALTER DATABASE flex`. So this should be safe.

db/flex/service_providing_group_grid_prequalification_sync.sql
```diff
@@ -0,0 +1,71 @@
+--liquibase formatted sql
+-- Manually managed file
+
+-- changeset flex:service-providing-group-grid-prequalification-sync-view runOnChange:true endDelimiter:--
+-- (SPG, ISO) pairs for which a SPG-GP record should exist in the DB
+CREATE OR REPLACE VIEW flex.service_providing_group_grid_prequalification_v AS
```
krilor: Should we be explicit about security definer?

db/flex/service_providing_group_product_application_migrations.sql
```diff
@@ -222,3 +222,35 @@ DROP COLUMN IF EXISTS maximum_active_power;
...
+--preconditions onFail:MARK_RAN
+--precondition-sql-check expectedResult:0 SELECT COUNT(*) FROM pg_catalog.pg_constraint WHERE conname = 'spg_product_application_status_check'
+ALTER TABLE flex.service_providing_group_product_application
+DISABLE TRIGGER USER;
+
+ALTER TABLE flex.service_providing_group_product_application
+DROP CONSTRAINT service_providing_group_product_application_status_check;
+
+UPDATE flex.service_providing_group_product_application
+SET status = 'prequalification'
```
krilor, line 236: are we missing a migration for status `prequalification_pending`?

db/flex/service_providing_group_grid_prequalification_sync.sql
```diff
@@ -0,0 +1,71 @@
...
+    -- TODO: missing in the view -> [soft-]delete the SPG-GP?
+    -- WHEN NOT MATCHED BY SOURCE THEN
+    --     DELETE;
+END;
+$$;
+
+-- changeset flex:service-providing-group-grid-prequalification-sync-job runAlways:true endDelimiter:;
+SELECT cron.schedule(
+    'service-providing-group-grid-prequalification-sync',
+    '0 2 * * *', -- every night at 2 AM
```
krilor: NIT: Generally we should not run things at the top of the hour. Everyone does that. It is good practice IMHO to add some random minutes and seconds (still hardcoded tho), just to avoid that we do this at the same time as external clients decide to sync.

frontend/src/dashboard/hooks/useDashboardApplications.ts
```diff
@@ -17,7 +17,7 @@ export const ACTIVE_STATUSES = new Set([
   "requested",
   "in_progress",
```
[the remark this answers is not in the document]
enzo-crance-statnett: Can be a status for SPPA for example, this is not only SPGPA things. Invalid comment.

docs/resources/service_providing_group_product_application.md
```diff
@@ -10,10 +10,16 @@ applications. This resource supports the application process and is updated as
 it progresses.
 
 The service provider is responsible for creating a new SPG product application.
-Initially, the status is set to `requested`. The system operator then takes over
-until the application is either `prequalified`, `verified` or `rejected`. If it
-rejected, the service provider can make updates and propose the application
-again.
+Initially, the status is set to `requested`. The system operator then has 4
```
krilor: I think we should be a bit careful with explicitly saying 4 weeks. We should maybe just say that the first step is to choose the process.

The NCDR wording is typically "without undue delay" 🤓

krilor (approved):
💯

PR #840, opened 2026-05-15
fix: forward label prop in Datagrid column headers (FLEX-1199)

s0lvang:
## Summary

- The custom `Datagrid` component in `EDS-ra/list/Datagrid.tsx` was only extracting the `source` prop from child column definitions, ignoring the `label` prop entirely
- This caused columns like the SPG prequalification fields — which use `label=` but no `source` — to render empty column headers
- Fixed by also extracting `label` from child props and passing it through to `<FieldTitle>`

Closes FLEX-1199


docs/superpowers/specs/2026-05-15-tr-empty-state-design.md
enzo-crance-statnett, line 1: Maybe we need to update gitignore or something

enzo-crance-statnett (dismissed):

PR #847, opened 2026-05-19
feat: adjusted field descriptions and form layout on TR (FLEX-1197)

krilor:
This PR mainly suggests two things to fix the problem of MAP being confusing on TR.

1. Adjust description text to be more clear 🤞
2. Change the layout of the form to move MAP in the middle of a new "device" section that highlights that this is about the device itself.

Hopefully this will help a little bit in avoiding confusion.

This is what it looks like:

bilde


test/flex/api/accounting_point/list_accounting_point.py
enzo-crance-statnett, line 1: I hope another PR does not revert this 😅 IIRC we have had this problem recently

enzo-crance-statnett (approved):

s0lvang (approved):

test/flex/api/accounting_point/list_accounting_point.py
krilor: 🤞A new version of the openapi client generated landed in main today. Was thinking it was related to that..

PR #852, opened 2026-05-20
feat(auth): create entity if not exists on successful IDP login (FLEX-1221)

enzo-crance-statnett:
This PR adds changes to the `/callback` endpoint in the auth API, so that a successful login in the IDP always causes an entity to exist for the user, so that they can get to the assume party page instead of getting a little user-friendly error message and having to contact their ORG admin to create their entity.

The auth mechanism translating an entity business ID into a technical ID now creates the entity if it does not already exist. All the call chain is updated. The name of the new entity is taken from the token, with various strategies based on the IDP, always falling back on the business ID itself on use of an unknown OIDC issuer.

One user without entity in the system has been added to the local config of Authelia so that we can test logging in and seeing the new entity created.

### DIFF
--- a/db/auth/functions.sql
+++ b/db/auth/functions.sql
@@ -27,22 +27,48 @@ $$;
 GRANT EXECUTE ON FUNCTION auth.entity_of_credentials TO flex_anonymous;
 
 -- changeset flex:auth-entity-of-business-id runAlways:true endDelimiter:--
--- Gets entity details from the business id
-CREATE OR REPLACE FUNCTION auth.entity_of_business_id(
+-- Gets entity details from the business id, creating it if it does not exist.
+DROP FUNCTION IF EXISTS auth.entity_of_business_id(text, text);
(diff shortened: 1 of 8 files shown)


db/auth/functions.sql
```diff
@@ -27,22 +27,48 @@ $$;
 GRANT EXECUTE ON FUNCTION auth.entity_of_credentials TO flex_anonymous;
 
 -- changeset flex:auth-entity-of-business-id runAlways:true endDelimiter:--
--- Gets entity details from the business id
-CREATE OR REPLACE FUNCTION auth.entity_of_business_id(
+-- Gets entity details from the business id, creating it if it does not exist.
+DROP FUNCTION IF EXISTS auth.entity_of_business_id(text, text);
```
krilor, line 30: NIT: Drop should maybe be it s own changeset since its for the old function, not the new/current one.

krilor (commented):
Sweet!

krilor (dismissed):

PR #858, opened 2026-05-21
feat(ui): improve entity lookup flow (FLEX-1223)

enzo-crance-statnett:
This PR merges several steps of adding a user to a party via entity lookup, into one modal never leaving the same page.

The user first inputs the entity business ID and name to run lookup with a button, then if the lookup succeeds an entity is retrieved or created and its technical ID is stored in the component. Then, they "unlock" the rest of the form where they just have to select the scopes. Submitting the form creates the party membership and the user has never left the party page.

The `ScopesInput` component could not be used as the modal is EDS-only and it caused z-index issues showing the dropdown list behind the modal. As scopes are about to change (with support for more general scopes), I did not want to invest time in making the perfect component there and replace all occurrences.

---

Starting from the list

Screenshot 2026-05-21 142152

First part of the form

Screenshot 2026-05-21 151412

When the entity is found/created

Screenshot 2026-05-21 155236


enzo-crance-statnett:
ref `ScopesInput`
It can be considered as a part of https://elhub.atlassian.net/browse/FLEX-1175 to improve/move this component, for example...

frontend/src/party/membership/PartyMembershipList.tsx
```diff
@@ -18,62 +20,61 @@ import { Button } from "../../components/ui";
...
 }) => {
   const canCreate = permissions?.allow("party_membership", "create");
   const canLookup = permissions?.allow("entity", "lookup");
 
   return (
     <>
-      {canCreate && <CreateButton id={id} />}
-      {canCreate && canLookup && <CreateViaLookupButton id={id} />}
+      {canCreate && isFISO && <CreateButton id={id} />}
+      {canCreate && canLookup && (
```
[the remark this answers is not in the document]
enzo-crance-statnett: The only other role with create on PM is ORG and they have the lookup button. I don't know to what extent we need to keep the create button at all here.

frontend/src/party/membership/AddPartyMemberViaEntityLookupModal.tsx
```diff
@@ -0,0 +1,281 @@
...
+import { createPartyMembership } from "../../generated-client";
+import { AuthScope } from "../../generated-client/types.gen";
+import { throwOnError } from "../../util";
+import { Identifier } from "ra-core";
+
+// ScopesInput
+// TODO: improve/move this component when we have support for general scopes
+// (this is reimplemented here to avoid z-index issues when mixing RA with EDS)
+
+const ALL_SCOPES: AuthScope[] = [
```
krilor: This should be an enum already in the generated models?

krilor (dismissed):
The flow looks nice and it is a good idea to make it more seemless like this. But we are not really adressing some the major pain points that I observe in the flow.

1. This is the biggest one: Professional users in envs running Entra are idenfied with email, not pid!
2. The user probably does not know what party memberships are. We should tell them what it is. More guiding text.
3. The user does not know what "lookup" is and it should not be a concept they need to think about. Just do "Add new".

frontend/src/party/membership/AddPartyMemberViaEntityLookupModal.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+    try {
+      const response = await fetch(apiURL + "/entity/lookup", {
+        method: "POST",
+        headers: new Headers({
+          "Content-Type": "application/json",
+          "Api-Version": API_VERSION,
+        }),
+        body: JSON.stringify({ business_id: businessId, name, type: "person" }),
+      });
```
s0lvang: Should probably use the generated client

s0lvang: And use the useMutation hook from React-query, then you dont have to handle error-states, loading etc

You can also use the onSuccess and on Error callbakcs

frontend/src/party/membership/AddPartyMemberViaEntityLookupModal.tsx
```diff
@@ -0,0 +1,281 @@
...
+      handleClose();
+    } catch (e) {
+      setSubmitError(
+        e instanceof Error
+          ? e.message
+          : "Failed to add member. Please try again.",
+      );
+      setPhase("entity_found");
+    }
+  };
```
s0lvang: useMutation here as well. Then you can remove the submitting phase because that is just isPending on the submit-mutation

frontend/src/party/membership/AddPartyMemberViaEntityLookupModal.tsx
```diff
@@ -0,0 +1,281 @@
...
+import { IconSearch, IconPlus } from "@elhub/ds-icons";
+import { apiURL, API_VERSION } from "../../httpConfig";
+import { createPartyMembership } from "../../generated-client";
+import { AuthScope } from "../../generated-client/types.gen";
+import { throwOnError } from "../../util";
+import { Identifier } from "ra-core";
+
+// ScopesInput
+// TODO: improve/move this component when we have support for general scopes
+// (this is reimplemented here to avoid z-index issues when mixing RA with EDS)
```
s0lvang, line 30: What are these z-index issues?

frontend/src/party/membership/AddPartyMemberViaEntityLookupModal.tsx
```diff
@@ -0,0 +1,281 @@
...
+export const AddPartyMemberViaEntityLookupModal = ({
+  partyId,
+  open,
+  onClose,
+}: Props) => {
+  const queryClient = useQueryClient();
+
+  const [phase, setPhase] = useState<Phase>("start");
+  const [businessId, setBusinessId] = useState("");
+  const [name, setName] = useState("");
```
s0lvang: We can use a form for this as well so we get input validation. zEntityLookupRequest is the schema

frontend/src/party/membership/AddPartyMemberViaEntityLookupModal.tsx
```diff
@@ -0,0 +1,220 @@
...
+  return (
+    <Modal open={open} onClose={handleClose} aria-label="Add member via lookup">
+      <Modal.Header
+        title="Add member via lookup"
+        description="Look up an entity by their personnummer or email address and name to add them as a member of this party."
+      />
+      <Modal.Content className="flex flex-col gap-4 min-w-lg">
+        <FormItem>
+          <FormItemLabel htmlFor="lookup-business-id">
+            Business ID (personnummer or email)
```
krilor: The UI should just be about email. Let's avoid encuraging the use of pid - it is from a time where we thought that professional users also will be identified using pid.  It is not the case any more. We should validate that the input contains a `@`.

frontend/src/party/membership/AddPartyMemberViaEntityLookupModal.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const errorMessage = mutationError
+    ? mutationError instanceof Error
+      ? mutationError.message
+      : "An unexpected error occurred. Please try again."
+    : null;
+
+  return (
+    <Modal open={open} onClose={handleClose} aria-label="Add member via lookup">
+      <Modal.Header
+        title="Add member via lookup"
```
krilor: Lets remove mentions of lookup in the UI?
Lookup is a technical step, not something that the user actually cares about. It should just be "under the hood" IMHO

frontend/src/party/PartyShow.tsx
```diff
@@ -50,8 +50,13 @@ export const PartyShow = () => {
       {!isHistory && (
         <>
           <Heading level={2} size="small" spacing>
-            Party memberships
+            Party members
           </Heading>
+          <BodyText>
+            The following users are allowed to assume this party in the system.
+            If you are an organisation administrator, you can add more members
+            to this party by clicking the button on the right.
```
krilor, line 58: 👍

backend/data/data.go
```diff
@@ -459,19 +459,45 @@ func (data *api) entityLookupHandler(
...
-			regexPersonBusinessID.MatchString(entityLookupRequestBody.BusinessID),
-			"person number must be 11 digit long",
+			expectedType == entityLookupRequestBody.Type,
+			"%s", "business_id_type '"+entityLookupRequestBody.BusinessIDType+
+				"' is not valid for entity type '"+entityLookupRequestBody.Type+"'",
 		)
 	}
 
+	regexPersonBusinessID := regexp.MustCompile("^[1-9][0-9]{10}$")
+	regexEmailBusinessID := regexp.MustCompile(`^[a-zA-Z0-9._%-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$`)
```
krilor: email validation using regex is famously hard. Ref https://stackoverflow.com/questions/201323/how-can-i-validate-an-email-address-using-a-regular-expression

Should we maybe just pass it to https://pkg.go.dev/net/mail#AddressParser and check for an error/not?

frontend/src/party/membership/AddPartyMemberViaEntityLookupModal.tsx
enzo-crance-statnett: Scope dropdown showing behind the modal

PR #859, opened 2026-05-22
fix(ui): better label generation on table headers (FLEX-1225)

enzo-crance-statnett:
This PR fixes a bug in labels in table headers, where we had stuff like `Business` instead of `Business ID`.

When there was no label, the name of the field was sent to the default translator in React-Admin, and we got back poorly informative headers.

Instead now we try to go through our field names in the resources YAML, and reference fields include the name of the referenced resource, unless `label` is given.

---

Some examples:

CU list after fix (`Business ID`, and the AP suffix to tell one ID from the other)

Screenshot 2026-05-22 151648

SPG list (first `Name` is SPG name, second `Name` is now clearly the SP's name, whereas before we had `Name` twice)

Screenshot 2026-05-22 161557

### DIFF
--- a/frontend/src/components/EDS-ra/list/Datagrid.tsx
+++ b/frontend/src/components/EDS-ra/list/Datagrid.tsx
@@ -4,6 +4,7 @@ import React, {
   cloneElement,
   isValidElement,
 } from "react";
+import { humanize } from "inflection";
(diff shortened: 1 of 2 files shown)


krilor (approved):
🤓

PR #861, opened 2026-05-23
feat: surface error messages in spg manage members (FLEX-1227)

krilor:
Allows the user to expand and view the error message.

bilde

Does not show as default.

### DIFF
--- a/frontend/src/service_providing_group/membership/FindControllableUnits.tsx
+++ b/frontend/src/service_providing_group/membership/FindControllableUnits.tsx
@@ -11,6 +11,7 @@ import {
   Loader,
   Modal,
 } from "../../components/ui";
+import { ErrorMessage } from "../../generated-client";
 import { IconPlus } from "@elhub/ds-icons";
 import {
   useAddMemberships,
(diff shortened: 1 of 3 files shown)


frontend/src/service_providing_group/membership/FindControllableUnits.tsx
```diff
@@ -82,10 +93,20 @@ export const FindControllableUnits = ({
   };
 
   const columns: ColumnOf<typeof availableCus>[] = [
+    {
+      key: "id",
+      header: t("controllable_unit.id"),
+      render: (v) => <span className="whitespace-nowrap">{String(v)}</span>,
```
s0lvang, line 99: Is it neccesary to have whitespace-nowrap on all columns?

s0lvang (approved):

frontend/src/service_providing_group/membership/FindControllableUnits.tsx
krilor: Not sure. I just added to the ones that was messed up in the UI. Should prob revisit. Maybe it should be a column prop or something instead.

PR #867, opened 2026-05-27
refactor: migrate show pages (FLEX-1210)

s0lvang:
migrates the remaining show pages to eds-ra. Some of these should probably just be modals as we do with TR but for the simplicity of the migration i kept them as is.


kenneth-strand-statnett:
Closing this as it seems to already be in main

PR #874, opened 2026-05-29
feat: nullable location on accounting point resource (FLEX-1133)

enzo-crance-statnett:
This PR adds a `location` field to the accounting point resource, as a PostGIS geometry (point) value. Through PostgREST it is surfaced as a GeoJSON point object in the API. We show the location as a map on one of the tabs in the accounting point show page.

Screenshot 2026-06-05 181815

We had to add the PostGIS extension to the local database and OpenFreeMap in the frontend. Since the map is downloading its style, we needed to update the content security policy. All this needs to be done on deployed environments as well.


db/flex/accounting_point.sql
```diff
@@ -9,8 +9,12 @@ CREATE TABLE IF NOT EXISTS accounting_point (
     business_id text UNIQUE NOT NULL CHECK (
         validate_business_id(business_id, 'gsrn')
     ),
+    latitude double precision NULL,
```
krilor: I really think we should store geographical points a PostGIS datatypes. The benefit is that we can do all kinds of geospatioal calculations with it - and we want to do that, e.g. finding distances or guessing substations.

See e.g.

* https://ncona.com/2025/05/postgis-working-with-geospatial-data-in-postgresql/
* https://postgis.net/documentation/faq/geometry-or-geography/

So, something like:

```
location geometry(point, 4326),
```

For exposing it to the world, I'm not sure what makes most sense. Either split it up into lat/long using ST_X/Y or expose it as geojson with ST_AsGeoJSON.

BTW: Here are some thoughts about GeoJSON related to gridnodes/substations: https://github.com/elhub/flex-information-system/blob/main/docs/technical/accounting-point-grid-location.md#geojson-geometry

krilor: Maybe geojson format output of the field is the way to go. It is automagic from postgrest https://docs.postgrest.org/en/stable/integrations/postgis.html

If we do that we can copy-pasta schema definitions (into reusable schemas) from https://github.com/geojson/schema

backend/data/static/openapi.json
```diff
@@ -4268,6 +4268,64 @@
...
                         "readOnly": true
                     },
+                    "location": {
+                        "description": "Geographic location of the accounting point (WGS84), as a GeoJSON point object.",
+                        "type": "object",
+                        "nullable": true,
+                        "readOnly": true,
+                        "example": {
+                            "type": "Point",
+                            "crs": {
```
krilor: hmmm...

https://datatracker.ietf.org/doc/html/rfc7946#appendix-B

> Changes from the Pre-IETF GeoJSON Format Specification
>
>   This appendix briefly summarizes non-editorial changes from the 2008
>   specification [[GJ2008](https://datatracker.ietf.org/doc/html/rfc7946#ref-GJ2008)].
>
> [B.1](https://datatracker.ietf.org/doc/html/rfc7946#appendix-B.1).  Normative Changes
> 
>   o  Specification of coordinate reference systems has been removed,
>      i.e., **the "crs" member of [[GJ2008](https://datatracker.ietf.org/doc/html/rfc7946#ref-GJ2008)] is no longer used**.

backend/data/static/openapi.json
```diff
@@ -4268,6 +4268,64 @@
...
+                                "properties": {
+                                    "name": "EPSG:4326"
+                                }
+                            },
+                            "coordinates": [
+                                -2.0259056,
+                                48.6504504
+                            ]
+                        },
+                        "properties": {
```
krilor: I think we should make this a named schema in the base openapi file. It should make it reusable and and ensure nicer naming in generated client (I think)

db/flex/accounting_point_migrations.sql
```diff
@@ -0,0 +1,15 @@
...
+--preconditions onFail:MARK_RAN
+--precondition-sql-check expectedResult:0 SELECT COUNT(*) FROM information_schema.columns WHERE table_schema = 'flex' AND table_name = 'accounting_point' AND column_name = 'location';
+
+ALTER TABLE flex.accounting_point_history
+ADD COLUMN IF NOT EXISTS location GEOMETRY (POINT, 4326) NULL;
+
+ALTER TABLE flex.accounting_point
+ADD COLUMN IF NOT EXISTS location GEOMETRY (POINT, 4326) NULL;
+
+CREATE INDEX IF NOT EXISTS accounting_point_location_idx
```
krilor, line 17: Adding the index should be a separate changeset IMHO.

The field is either added from the CREATE TABLE statement _or_ the ALTER TABLE above.
The should be added regardless of where the field stems from - therefore its own thing.

Each changeset should have a distinct purpose.

frontend/src/accounting_point/show/AccountingPointShow.tsx
```diff
@@ -81,6 +81,7 @@ export const AccountingPointShow = () => {
       <AccountingPointShowTabs
         accountingPoint={ap}
         gridLocation={viewModel.gridLocation}
```
krilor, line 83: Should we make it one "location" tab?

At least I _think_ that we should make all-things-location part of the same view - it would make sense to me at least.

backend/data/static/openapi.json
krilor: Btw: Nullable in OAS 3.1 is a bit wacky when dealing with refs

https://stackoverflow.com/a/48114924

backend/data/static/openapi.json
krilor: Internally postgrest uses ST_AsGeoJSON

According to https://postgis.net/docs/ST_AsGeoJSON.html

> The resulting GeoJSON geometry and feature representations conform with the [GeoJSON specifications RFC 7946](https://tools.ietf.org/html/rfc7946), **except when the parsed geometries are referenced with a CRS other than WGS84** longitude and latitude ([EPSG:4326](https://epsg.io/4326), [urn:ogc:def:crs:OGC::CRS84](http://epsg.io/4326.gml)); the GeoJSON geometry object will then have a short CRS SRID identifier attached by default.

enzo-crance-statnett: I am basing the spec on what PostgREST actually returns.

Example of response in the local test environment:

`GET /accounting_point/2003`
```json
{
    "id": 2003,
    "business_id": "133700000000020020",
    "system_operator_id": 26,
    "location": {
        "type": "Point",
        "crs": {
            "type": "name",
            "properties": {
                "name": "EPSG:4326"
            }
        },
        "coordinates": [
            5.3234103,
            60.3972037
        ]
    },
    "recorded_by": 0,
    "recorded_at": "2026-06-08T16:17:49.705445+00:00"
}
```

If we were to design the resource entirely by ourselves, I would just return the coordinates for this `location` field for simplicity's sake, but since we are going through a tool, I preferred the simple solution of not reformatting what it gives us and just documenting it transparently.

backend/data/static/openapi.json
enzo-crance-statnett: I got nicer naming, but it was a bit of a challenge to make both frontend types, Python test client and Kotlin models happy. Ended up adding an `x-*` temporary property to ease codegen.

frontend/src/accounting_point/show/AccountingPointShow.tsx
enzo-crance-statnett: With merged tabs

Screenshot 2026-06-11 142925

db/api/accounting_point.sql
```diff
@@ -2,13 +2,15 @@
 -- Manually managed file
 
 -- changeset flex:api-accounting-point-create endDelimiter:-- runOnChange:true
-CREATE OR REPLACE VIEW api.accounting_point
+DROP VIEW IF EXISTS api.accounting_point CASCADE;
+CREATE VIEW api.accounting_point
```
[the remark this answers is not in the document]
enzo-crance-statnett: Liquibase was not happy with changing the order of fields (nor was the SQL linter). Everything using this view is normally recreated afterwards, but it is true we now have `runOnChange` and not `runAlways` in this schema, so we need to be more careful. I think Copilot is right and we need to revert.

db/flex/accounting_point.sql
```diff
@@ -9,6 +9,7 @@ CREATE TABLE IF NOT EXISTS accounting_point (
     business_id text UNIQUE NOT NULL CHECK (
         validate_business_id(business_id, 'gsrn')
     ),
+    location GEOMETRY (POINT, 4326) NULL,
```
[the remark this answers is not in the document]
enzo-crance-statnett: Known and mentioned from the start in the PR description.

db/api/accounting_point.sql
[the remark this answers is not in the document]
krilor: I think we should never have CASCADE in our migrations. Scary stuff.

backend/data/static/openapi.json
krilor: Hmm.. I'll have a look at why crs is returned. I don't want it. I'm probably reading the docs wrong, but I thought it would not be there if we used the proper CRS internally.

backend/data/static/openapi.json
krilor: Hmm. The output should be something like this?
https://github.com/elhub/flex-information-system/blob/31c74539e8fe92be0c2cbb731ef0ad7e88434489/backend/data/static/openapi.json#L6622-L6634

backend/data/static/openapi.json
krilor: I went on a bit of a deep dive on this and it seems like whats is going on under the hood is that PostgREST is simply doing to_json, not using ST_AsGeoJSON when the content type is application/json. The PostGIS internal typecast from geometry to json uses a c function that includes the csr. So our best bet is to force the use of ST_AsGeoJSON. Luckily we maintain the ap api view manually, so we can just:

```
ST_AsGeoJSON(ap.location,9,0)::json AS location,
```

db/api/accounting_point.sql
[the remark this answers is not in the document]
krilor: I guess the right thing to do in this case is to add location as the last column and keep CREATE OR REPLACE.

backend/data/static/openapi.json
enzo-crance-statnett: Got the result by forcing it in `resources.yml`, leaving `type: object` and `nullable: true` for the usual processing to continue working.

krilor (dismissed):
Nice!

PR #875, opened 2026-05-29
docs: update-registration-guides

Brynhildur-Eddudottir-Statnett:


Brynhildur-Eddudottir-Statnett:
Couldn't access portal. Need to double check that it corresponds to the portal flow/information.

PR #884, opened 2026-06-01
feat(ui): sensitive info warning on additional information fields (FLEX-1247)

enzo-crance-statnett:
This PR adds possibility to write additional information when creating a new CU, and adds a warning to all such fields across the app to prevent users from writing sensitive information there.

The technical solution for the latter change is factored into a `warning` prop on input components, so we can generalise this later if needed.

Screenshot 2026-06-01 162659


krilor:
Nice. The three categories of information we want to avoid is:

* power-sensitive information
* market-sensitive information
* special categories of personal data / sensitive personal data

Maybe the warning should be something like

>  Do not write any sensitive (power / market / personal) information.


Or along those lines?

PR #885, opened 2026-06-02
Docs: Update SPG registration process

Brynhildur-Eddudottir-Statnett:


docs/diagrams/service_providing_group_registration.plantuml
```diff
@@ -35,35 +34,38 @@ activate SP
 
 == Add Controllable Units (CU) to Service Providing Group (SPG) ==
 
-SP -> FIS: List own CU's with input filters
+SP -> FIS: Choose CU's from list of CU's
 activate FIS
-FIS --> SP: Return list of CU's
```
krilor, line 40: This step is still there in the process.

We must remember [this distinction between process and UI](https://elhub.github.io/flex-information-system/processes/)

> The process descriptions are a generalized view of how interactions are done between parties and systems in the value chain. It does not show the actual user experience, but rather logical steps that must be performed, e.g. towards the FIS API.
>
> It is up to the FIS UI and market party backend system to provide a good user experience on top. Some interactions might be fully automated between systems, some might be chained together from a single user interface and some cases might require the user to follow the process steps quite literally.

PR #899, opened 2026-06-04
fix(backend): midnight-align CU lookup validFrom on Norwegian timezone (FLEX-1262)

enzo-crance-statnett:
This PR fixes a bug that we had in CU lookup, where we were taking the _system midnight_ as the lower bound of the time window when asking the adapter for data, instead of the _Norwegian midnight_.

The bug has been silent because of the previous logic we had when filtering the timeline in the adapter, where any record _overlapping_ with the time window was kept. Then, unless a record was starting exactly on the asked start date (the wrong one -- so it never happened), the earliest record was kept with its start date even though it is earlier than the (wrong) limit we send. So the AP sync process worked and did not complain. With the new filtering, all records' start dates get replaced with the (wrong) `validFrom` if they happen to be earlier, so the DB refuses to upsert (the bug is made visible).

Fixed by explicitly using Norwegian timezone everywhere. Added an extra test as well.

### DIFF
--- a/kbackend/src/main/kotlin/no/elhub/flex/Application.kt
+++ b/kbackend/src/main/kotlin/no/elhub/flex/Application.kt
@@ -3,6 +3,7 @@ package no.elhub.flex
 import io.ktor.server.application.Application
 import io.ktor.server.application.install
 import io.ktor.server.netty.EngineMain
+import kotlinx.datetime.TimeZone
 import no.elhub.flex.auth.FlexAuthentication
 import no.elhub.flex.config.Tracing
 import no.elhub.flex.config.configureDatabase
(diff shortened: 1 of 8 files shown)


kenneth-strand-statnett (changes requested):
Instead of hardcoding timezone and using "norwegianMidnight" in the code, I believe it would be better to set the timezone for the application. Either as an environment variable `TZ=Europe/Oslo`, or with a JVM arg in the flex repository.

What do you think about this apporach?

enzo-crance-statnett:
In principle I agree, but we already hardcode this timezone in many other places (DB, frontend, tests). I'll try to fix this in a quick way, but doing a thorough fix across the app is out of scope.

kenneth-strand-statnett (approved):

PR #905, opened 2026-06-05
feat: Sync MGA data from adapter (FLEX-1263)

kenneth-strand-statnett:


db/flex/metering_grid_area_rls.sql
```diff
@@ -11,3 +11,11 @@ CREATE POLICY "MGA_COM001" ON metering_grid_area
 FOR SELECT
 TO flex_common
 USING (true);
+
+-- RLS: MGA-INT001
+GRANT INSERT, SELECT, UPDATE, DELETE ON metering_grid_area TO flex_internal_data;
```
krilor: the MGA table should should not be changed flex_internal_data. The data should only be managed by the structure data loading via staging.

krilor: ..I guess "SELECT" is the only thing needed on metering_grid_area

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointMeteringGridAreaRepository.kt
```diff
@@ -0,0 +1,82 @@
...
+                        apmga.accounting_point_id = src.accounting_point_id
+                        AND lower(apmga.valid_time_range) = src.valid_from
+                    )
+                    WHEN MATCHED AND (
+                        apmga.metering_grid_area_id IS DISTINCT FROM src.metering_grid_area_id
+                        OR upper(apmga.valid_time_range) IS DISTINCT FROM src.valid_to
+                    ) THEN UPDATE SET
+                        metering_grid_area_id = src.metering_grid_area_id,
+                        valid_time_range      = tstzrange(src.valid_from, src.valid_to, '[)')
+                    WHEN NOT MATCHED
```
krilor, line 63: We seem to be missing a delete-in-target as part of this sync. I don't think we need to handle that in this PR, but we should do a follow-up - possibly using the new PG17 when not mached stuff

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointMeteringGridAreaRepository.kt
enzo-crance-statnett, line 1: Why is this a distinct repository file, whereas the other AP-related data synchronisations happen in the common `AccountingPointRepository.kt`?

I think this is confusing to introduce a new pattern for just one of the syncs and not the others. But maybe there is a specific reason with the MGA one that I did not get.

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/AccountingPointService.kt
```diff
@@ -104,6 +108,25 @@ class AccountingPointServiceImpl(
...
+                                    accountingPointId = accountingPointId,
+                                    meteringGridAreaId = mgaMap.getValue(adapterMga.businessId).id,
+                                    validFrom = adapterMga.validFrom,
+                                    validTo = adapterMga.validTo,
+                                )
+                            }
+                            accountingPointMeteringGridAreaRepository.upsertAll(accountingPointMgas)
+                                .mapLeft { err -> err.toInternalServerError("upsertAll MGAs") }
+                                .bind()
+
```
enzo-crance-statnett, line 129: For example I was expecting
```kotlin
val meteringGridAreas = adapterAccountingPoint.toAccountingPointMeteringGridAreas(accountingPointId)
```
and then
```kotlin
accountingPointRepository.upsertAccountingPointMeteringGridAreas(meteringGridAreas)
```
like we did with the 2 other resources (AP-EU / AP-ES). They also have logic to retrieve the technical IDs from the business IDs (GLN -> party ID), this is similar with just EIC -> MGA ID instead. It allows for a less messy "main sync" function with implementation details hidden behind a dedicated function.

enzo-crance-statnett (commented):
Looks correct, I just have one question related to patterns/consistency across the code.

PR #913, opened 2026-06-08
feat: reintroduce controllable unit lookup events (FLEX-1265)

enzo-crance-statnett:
This PR reintroduces the insertion of events at the end of the CU lookup handler, which probably got removed when switching to the Kotlin backend.

The new events, however, use another logic than the one previously used: they use the _accounting point_ as the `source` instead of the controllable unit, which means only one event is generated per lookup. When it is a lookup by controllable unit business ID, we store this information in the `subject` field, and when it is a lookup by accounting point business ID, the `subject` is empty. All the previous logic generating notification should be preserved, it just needs to go through the AP ID instead of the CU ID in some cases.

Pre-existing events with the AP in `source` are updated to match the new format, even though they all probably have already been processed.

### DIFF
--- a/backend/event/models.sql
+++ b/backend/event/models.sql
@@ -334,12 +334,10 @@ WHERE cueu.controllable_unit_id = @cu_id
     AND cueu.valid_time_range @> @recorded_at::timestamptz;
 
 -- name: GetControllableUnitLookupNotificationRecipientsAP :many
-SELECT cueu.end_user_id
-FROM api.controllable_unit AS cu
-    INNER JOIN notification.controllable_unit_end_user AS cueu
-        ON cu.id = cueu.controllable_unit_id
-            AND cueu.valid_time_range @> @recorded_at::timestamptz
-WHERE cu.accounting_point_id = @ap_id;
+SELECT apeu.end_user_id
+FROM api.accounting_point_end_user AS apeu
+WHERE apeu.accounting_point_id = @ap_id
+    AND apeu.valid_time_range @> @recorded_at::timestamptz;
 
 -- name: GetServiceProvidingGroupGridSuspensionNotificationRecipients :many
 -- SP
--- a/backend/event/models.sql
+++ b/backend/event/models.sql
@@ -327,11 +327,19 @@ INSERT INTO api.notification (event_id, party_id)
 SELECT @event_id, unnest(@party_ids::bigint[])
 ON CONFLICT DO NOTHING;
 
--- name: GetControllableUnitLookupNotificationRecipients :many
+-- name: GetControllableUnitLookupNotificationRecipientsCU :many
 SELECT cueu.end_user_id
 FROM notification.controllable_unit_end_user AS cueu
-WHERE cueu.controllable_unit_id = @resource_id
+WHERE cueu.controllable_unit_id = @cu_id
     AND cueu.valid_time_range @> @recorded_at::timestamptz;
+
+-- name: GetControllableUnitLookupNotificationRecipientsAP :many
+SELECT cueu.end_user_id
+FROM api.controllable_unit AS cu
+    INNER JOIN notification.controllable_unit_end_user AS cueu
+        ON cu.id = cueu.controllable_unit_id
+            AND cueu.valid_time_range @> @recorded_at::timestamptz
+WHERE cu.accounting_point_id = @ap_id;
 
 -- name: GetServiceProvidingGroupGridSuspensionNotificationRecipients :many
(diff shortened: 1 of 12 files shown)


backend/event/models.sql
```diff
@@ -327,12 +327,20 @@ INSERT INTO api.notification (event_id, party_id)
...
+-- name: GetControllableUnitLookupNotificationRecipientsCU :many
 SELECT cueu.end_user_id
 FROM notification.controllable_unit_end_user AS cueu
-WHERE cueu.controllable_unit_id = @resource_id
+WHERE cueu.controllable_unit_id = @cu_id
     AND cueu.valid_time_range @> @recorded_at::timestamptz;
 
+-- name: GetControllableUnitLookupNotificationRecipientsAP :many
+SELECT cueu.end_user_id
+FROM api.controllable_unit AS cu
```
krilor: Shouldn't this query use the accounting_point and accounting_point_end_user table?
It feels a bit "weird" to use the CU tables

backend/event/models.sql
```diff
@@ -327,12 +327,20 @@ INSERT INTO api.notification (event_id, party_id)
...
--- name: GetControllableUnitLookupNotificationRecipients :many
+-- name: GetControllableUnitLookupNotificationRecipientsCU :many
 SELECT cueu.end_user_id
 FROM notification.controllable_unit_end_user AS cueu
-WHERE cueu.controllable_unit_id = @resource_id
+WHERE cueu.controllable_unit_id = @cu_id
     AND cueu.valid_time_range @> @recorded_at::timestamptz;
 
+-- name: GetControllableUnitLookupNotificationRecipientsAP :many
+SELECT cueu.end_user_id
```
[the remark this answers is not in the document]
krilor: We should apply this suggestion if we are not moving to using the ap tables instead.

kbackend/src/main/kotlin/no/elhub/flex/event/db/EventRepository.kt
```diff
@@ -0,0 +1,97 @@
...
+                // general lookup on the AP: no subject
+                conn.prepareNamed(
+                    """
+                INSERT INTO flex.event (type, source_resource, source_id, data)
+                SELECT
+                    public.text2ltree('no.elhub.flex.controllable_unit.lookup'),
+                    'accounting_point',
+                    (SELECT id FROM flex.accounting_point WHERE business_id::text = :apBusinessId),
+                    jsonb_build_object(
+                        'requesting_party_name',
```
krilor: Should we get rid of this "requesting_party_name" field? Does it really make sense to take a point-in-time name of a party part as a data field?

kbackend/src/main/kotlin/no/elhub/flex/routes/controllableunit/ControllableUnitLookup.kt
```diff
@@ -74,6 +78,12 @@ class ControllableUnitLookup(
                 ).bind()
                 val accountingPoint = accountingPointService.getAccountingPointByBusinessId(accountingPointBusinessId).bind()
 
+                insertEvent(
+                    accountingPointBusinessId,
```
krilor: We already know the internal IDs. Shouldn't we just pass IDs?

kbackend/src/main/kotlin/no/elhub/flex/event/db/EventRepository.kt
```diff
@@ -0,0 +1,97 @@
...
+    ): Either<RepositoryError, Unit>
+}
+
+private val logger = KotlinLogging.logger {}
+
+@Single(createdAtStart = true)
+class EventRepositoryImpl : EventRepository {
+
+    context(principal: FlexPrincipal)
+    override suspend fun insertLookupEvent(
```
krilor: Does it make sense that a eventRepository contains business logic like this?
Wouldn't it make more sense that the calling code just sends source/subject (incl ids) and the repository is just about storing it?

kbackend/src/main/kotlin/no/elhub/flex/event/db/EventRepository.kt
enzo-crance-statnett: Used the technical `id` instead.

backend/event/models.sql
enzo-crance-statnett: Yes it will be the same end user anyway because CU-EU gets its data from AP-EU. Switched.

krilor (approved):

PR #936, opened 2026-06-16
feat: scheduled job that syncs accounting point data (FLEX-1267)

kenneth-strand-statnett:


kbackend/src/main/kotlin/no/elhub/flex/scheduled/AccountingPointSyncScheduler.kt
```diff
@@ -0,0 +1,70 @@
...
+@Single(createdAtStart = true)
+class AccountingPointSyncScheduler(
+    private val accountingPointSyncRepository: AccountingPointSyncRepository,
+    private val accountingPointService: AccountingPointService,
+    private val controllableUnitRepository: ControllableUnitRepository,
+    @Property("flex.timezone") private val timezone: TimeZone = TimeZone.of("Europe/Oslo"),
+) {
+    suspend fun start() {
+        val batchSize = 50
+        doInfinity("* */5 * * *") { runBatch(batchSize) }
```
kenneth-strand-statnett: I've simplified the suggested calculation of batch size in the docs. The current config will let us process ~14k every day per instance.

docs/technical/accounting-point-data.md
```diff
@@ -163,14 +163,8 @@ The following participants are involved in the flows:
...
 
-The batch size for the background sync can be established upon application
-startup. We generally want all accounting points to be synced within 24 hours.
-The batch size should be configured as follows. We can use startup since
-we assume that we will deploy/recycle the application quite often.
-
-```none
-Batch size = nb_accounting_points / ((24h / interval) * nb_workers)
-```
+The batch size for the background sync is set to a static size of 50.
```
krilor, line 166: Good simplification. Just need to remember that we have this debt when when/if we cross the magic 14k * instances limit.

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointSyncRepository.kt
```diff
@@ -0,0 +1,51 @@
...
+    context(principal: FlexPrincipal)
+    override suspend fun getBatchForSync(batchSize: Int): Either<RepositoryError, List<Long>> = flexTransaction { conn ->
+        Either.catch {
+            conn.prepareNamed(
+                """
+                WITH batch AS (
+                    SELECT accounting_point_id
+                    FROM flex.accounting_point_sync
+                    WHERE (last_synced_at IS NULL OR last_synced_at < NOW() - INTERVAL '23 hours')
+                      AND (last_sync_start IS NULL OR last_sync_start < NOW() - INTERVAL '5 minutes')
```
krilor: The 5 minutes is there to detect dead syncs, I guess. Are we sure that it is dead by then?

Also should consider adding a why-comment to justify why 5 mins is enough

kbackend/src/main/kotlin/no/elhub/flex/config/Scheduling.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import io.github.oshai.kotlinlogging.KotlinLogging
+import io.ktor.server.application.Application
+import kotlinx.coroutines.launch
+import no.elhub.flex.scheduled.AccountingPointSyncScheduler
+import org.koin.ktor.ext.get
+
+private val logger = KotlinLogging.logger {}
+
+fun Application.configureScheduling() {
```
krilor, line 12: How does this handle termination?
My kotlinsk is weak, but I would assume that we need to propagate/handle SIGTERM?

kbackend/src/main/kotlin/no/elhub/flex/scheduled/AccountingPointSyncScheduler.kt
```diff
@@ -0,0 +1,67 @@
...
+                }
+
+                logger.info { "Processing batch of ${accountingPointIdBatch.size} accounting points" }
+
+                val accountingPoints = accountingPointService.getByIds(accountingPointIdBatch).bind()
+
+                accountingPoints.forEach { accountingPoint ->
+                    either {
+                        logger.debug { "Syncing accounting point with id ${accountingPoint.id}.." }
+                        val controllableUnits = controllableUnitRepository.getByAccountingPointId(accountingPoint.id).bind()
```
krilor: Wouldn't it be benefitial to join the data and pick out start date when looking up accouting points by IDs? Why do we want to do it in a loop in application code?

kbackend/gradle/libs.versions.toml
```diff
@@ -80,6 +80,7 @@ ktor-client-content-negotiation = { group = "io.ktor", name = "ktor-client-conte
 ktor-server-netty = { group = "io.ktor", name = "ktor-server-netty", version.ref = "ktor" }
 ktor-server-swagger = { group = "io.ktor", name = "ktor-server-swagger", version.ref = "ktor" }
 ktor-server-status-pages = { group = "io.ktor", name = "ktor-server-status-pages", version.ref = "ktor" }
+krontab = { group = "dev.inmo", name = "krontab", version.ref = "krontab" }
```
krilor, line 83: NIT: We could avoid this dependency by just running a bare while loop with a simple `dothething + wait/sleep 5ish-minute`

We don't really need to run this at exactly 5 minutes. To be honest, its nice for the service we call to not have it on the full minute - everyone does that. A sleep could be something like 286548 milliseconds to have a bit of "jitter"

kbackend/src/main/kotlin/no/elhub/flex/scheduled/AccountingPointSyncScheduler.kt
kenneth-strand-statnett: I did consider fetching a map of <AccountingPointId, LocalDate> before the loop, but I figured this way would make it easier to read the code. I'm open to changing it.

kbackend/src/main/kotlin/no/elhub/flex/config/Scheduling.kt
kenneth-strand-statnett: It does handle sigterm through Ktor application lifecycle. However, we might log that it stopped unexpectedly when this happens. Is that a problem? If yes, then we can try to not log on CancellationException.

kbackend/src/main/kotlin/no/elhub/flex/scheduled/AccountingPointSyncScheduler.kt
krilor: I think readability is a bit subjective here. I think we should get used to joining and filtering data in the database rather than doing it in application code. One thing I'm a bit curious about is how this fits with how we have structured repositories up until now. Like this is a join between AP and CU. What repository does it fit in?

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointSyncRepository.kt
kenneth-strand-statnett: This was just guessing on my part. I would be surprised if it takes more than a minute to process one batch.
Do you think we should increase it?

kbackend/gradle/libs.versions.toml
kenneth-strand-statnett: For this specific job I agree that we don't really need a scheduling library, but we might need it in the future (e.g. run at 01:00 every day), and I like the idea of having the same setup for all jobs.

I guess I'm just used to having it easily available in other frameworks 😅

kbackend/src/main/kotlin/no/elhub/flex/config/Scheduling.kt
krilor: Ok nice.

If it is an error log then I think it is a problem. We should follow up/alert on all error logs. If its logged on a info level then less of a problem.

Ref https://github.com/elhub/flex-information-system/blob/main/docs-dev/logging.md#levels

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointSyncRepository.kt
krilor: Ah ok. I  think we should. We are working with "sync every day", so we can make this an hour to reduce the risk at least.

Another way is to set an explicit timeout on the batch processing to be lower than the interval here.

kbackend/src/main/kotlin/no/elhub/flex/scheduled/AccountingPointSyncScheduler.kt
kenneth-strand-statnett: I have no problems with joins and filtering in the database, but I am a bit allergic to APIs that reads like
`getLowestStartDateForAllControllableUnitsPerAccountingPoint(accountingPointIds: List<Long>): Map<Long, LocalDate>`, so if you have a suggestion for a name.. 🙏  😅 

 I'm used to dividing repositories by the main resource you're querying on, in this case I believe it would be the repository for controllable units. Basically `select min(start_date) from controllable_unit where accounting_point_id = ANY(:ids) GROUP BY accounting_point_id`

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointSyncRepository.kt
kenneth-strand-statnett: Aight, I'll set it to one hour. The code will time out within this limit anyways, either by http client timeout or database connect/read.

kbackend/src/main/kotlin/no/elhub/flex/config/Scheduling.kt
kenneth-strand-statnett: fixed by rethrowing instead of logging error for coroutine cancellations.

kbackend/src/main/kotlin/no/elhub/flex/scheduled/AccountingPointSyncScheduler.kt
enzo-crance-statnett: `getEarliestControllableUnitStartDates` with a line of documentation is acceptable IMO. Long names do not necessarily make things clearer.

enzo-crance-statnett (approved):

PR #943, opened 2026-06-18
feat: grid API and map picker for accounting point grid location (FLEX-1134)

enzo-crance-statnett:
This PR opens the grid data to a new `grid` module in the API, under a `/grid/v0` prefix. As it is still moving and we are not sure about what we want to expose and how we want it to work, I left it unspecified. We also need to fix various scripts so they work for several modules. The new module is solely based on PostgREST to keep it simple. I considered moving some PostgREST-specific functions to a common package in the backend, but the handler for the `grid` module is so much simpler (and therefore easy to move to kbackend if we want more advanced functionalities) that I did not want to invest this time now.

Then, we implement picking the grid location from the map. In order not to load the whole substation data to show the transformers on the map (3k lines, too heavy), I thought about a 10 km radius filter, IDK if it makes much sense functionally. It is necessary to implement this filter backend-side, but the PostgREST documentation does not seem to propose any special filter that would work on PostGIS data to have a query like `?dist=(id,10000)` or similar. So the filtering is implemented in a DB view, with a PostGIS function doing exactly what we need. Performance-wise this should be OK as we only call the view with an accounting point filter, but this is not enforced backend-side yet.

We only extract a list of substation IDs that are the near active transformers to an accounting point, the rest of the data is currently fetched with another query. I did this split because in the future I expect that we will use embedding like in the data API, so we will be able to simply delete the second query.

The grid endpoints are set behind the `data` scope asset to test things faster, but they should probably live under their own `grid` asset at some point. I just thought this would be a bit much to also update the scopes across the app in the same PR. But this is an important TODO because if we don't implement that, it's a surprise factor for the users.

Visually we show the accounting point's geographical location, the near transformers, the substation clusters they belong to, and the lines starting from them. IDK which of this information is relevant and which is unnecessary (we should keep the map as simple as possible). Colors are not the nicest but I needed something that is not hardcoded so I reused some of the default ones. This is more of a quick experiment about visualisation.

Hovering a transformer shows its name, clicking on it fills automatically the underlying form so that the user can update the grid location in very few steps. Internally we store the properties inside the GeoJSON features on the map, so that data can be filled directly from there.

---

Some screenshots

Screenshot 2026-06-18 164956

Hovering

Screenshot 2026-06-18 165003

Autofill of form

Screenshot 2026-06-18 162848

Selecting another transformator a bit further

Screenshot 2026-06-18 165123


enzo-crance-statnett:
Switched the endpoint to a RPC call so we can use a more general filter without a hardcoded distance.

A call to `GET /substation?order=proximity&longitude=<long>&latitude=<lat>` on the backend API becomes `GET /rpc/substation&longitude=<long>&latitude=<lat>` on the PostgREST API.

The function is exposed as a `GET` and arguments are given as query parameters, because it is marked `IMMUTABLE`. Makes it easier than having to translate to `POST` with a body. Using a `GET` means further filters are also automatically passed down to the underlying query. This is useful for instance in the frontend where we add filters to get active transformers only.

A limit is set to 10 if no limit is set or if clients set a higher limit, so we are sure we don't return too much data.

Calls to `GET /substation` without the `order=proximity` parameter are just forwarded to the general PostgREST list endpoint.

backend/grid/grid.go
```diff
@@ -73,9 +76,29 @@ func (grid *grid) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+		query.Del("order")
+
+		// target the function instead of the table, so we can use the geographical
+		// filtering mechanism
+		req.URL.Path = "/rpc/substation_distance"
+		req.URL.RawQuery = query.Encode()
 
+		// set a limit to max 10 if not set, so we do not return the whole dataset
+		limit, err := strconv.Atoi(query.Get("limit"))
+		if err != nil || limit > 10 {
```
krilor: Should we maybe allow getting more than 10? I agree to 10 if not set, but should we remove the max?

db/grid/substation_distance.sql
```diff
@@ -0,0 +1,30 @@
...
+-- Manually managed file
+
+-- changeset flex:grid-substation-distance endDelimiter:-- runOnChange:true
+-- substations ordered by distance to a given point
+CREATE OR REPLACE FUNCTION grid.substation_distance(
+    longitude double precision,
+    latitude double precision
+)
+RETURNS SETOF grid.substation
+SECURITY DEFINER
```
krilor: Should be security invoker

enzo-crance-statnett:
Regarding the `SECURITY DEFINER` typo, I revisited as discussed the grants mechanisms we have for functions in general, to avoid that a typo wrongly gives a lot of access to a user that is not meant to have it in the future.

The default privilege in Postgres is actually that [anyone can execute any function](https://www.postgresql.org/docs/current/ddl-priv.html#:~:text=PostgreSQL%20grants%20privileges,DEFAULT%20PRIVILEGES%20command.). This is not good for us.

I updated the defaults on all the concerned schemas, and revoked all these grants and gave them again only to the right roles. I hope this adds the security layer that we expect. Most of this is not exposed by PostgREST and only called internally, so in general we should not be too worried, but still.

This last change can probably be taken to another PR if we want. We'll just have to repeat the `grid`-specific part here.

frontend/src/accounting_point/show/AccountingPointLocationMap.tsx
```diff
@@ -1,15 +1,188 @@
...
+// TODO: replace with an API client generated from the spec
+
+const fetchJSON = async <T,>(url: string): Promise<T[]> => {
+  const response = await fetch(url, {
+    headers: { Accept: "application/json" },
+    credentials: "include",
+  });
+  if (!response.ok) return [];
+  return response.json();
+};
```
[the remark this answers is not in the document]
krilor: Makes sense to me.

db/grid/substation_distance.sql
```diff
@@ -0,0 +1,37 @@
...
+    substation_cluster_id,
+    voltage_levels,
+    recorded_by,
+    lower(record_time_range) AS recorded_at
+FROM flex.substation
+ORDER BY position <-> ST_SetSRID(ST_MakePoint(longitude, latitude), 4326);
+$$;
+
+-- changeset flex:grid-substation-distance-grants runOnChange:true
+REVOKE EXECUTE ON FUNCTION grid.substation_distance(
```
krilor: Don't think we need this since we create the function in the same change/PR.

backend/grid/grid.go
```diff
@@ -0,0 +1,227 @@
...
+	// grid schema only
+	header.Set("Accept-Profile", "grid")
+
+	// rewrite the URL and query to match the PostgREST format
+	if match := regexSingleID.FindStringSubmatch(url); match != nil {
+		url = "/" + match[1]
+		query.Set("id", "eq."+match[2])
+		// force PostgREST to return a single object
+		header.Set("Accept", "application/vnd.pgrst.object+json")
+		slog.InfoContext(
```
krilor: NIT: This should maybe be DebugContext

backend/grid/grid.go
```diff
@@ -0,0 +1,227 @@
...
+	mux.Handle("GET /substation_history", listPostgRESTHandler)
+	mux.Handle("GET /substation_history/{id}", postgRESTHandler)
+
+	mux.Handle("GET /substation_cluster", listPostgRESTHandler)
+	mux.Handle("GET /substation_cluster/{id}", postgRESTHandler)
+
+	mux.Handle("GET /substation_cluster_history", listPostgRESTHandler)
+	mux.Handle("GET /substation_cluster_history/{id}", postgRESTHandler)
+
+	mux.Handle("GET /accounting_point_near_transformer", listPostgRESTHandler)
```
krilor: Forgot to remove this

db/auth/functions.sql
```diff
@@ -271,3 +271,17 @@ $$;
 
 -- changeset flex:pre-request-execute runOnChange:true endDelimiter:;
 GRANT EXECUTE ON FUNCTION auth.pre_request TO flex_anonymous;
+
+-- changeset flex:auth-function-grants runOnChange:true endDelimiter:;
+REVOKE EXECUTE ON FUNCTION auth.entity_of_credentials(text, text) FROM public;
+REVOKE EXECUTE ON FUNCTION auth.get_or_create_entity(text, text, text) FROM public;
```
krilor: Should these live closer to the functions to align with the rest of the change and to simplify future maintenance?

kbackend/src/test/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointRepositoryTest.kt
```diff
@@ -26,7 +26,7 @@ class AccountingPointRepositoryTest : FunSpec({
 
     val repo = AccountingPointRepositoryImpl()
 
-    val principal = FlexPrincipal(role = "flex_flexibility_information_system_operator", eid = "0")
+    val principal = FlexPrincipal(role = "flex_internal_data", eid = "0")
```
krilor: 👍

backend/grid/grid.go
```diff
@@ -0,0 +1,225 @@
...
+
+	mux := http.NewServeMux()
+
+	grid := &grid{
+		postgRESTURL: postgRESTURL,
+		mux:          mux,
+	}
+
+	listPostgRESTHandler := middleware.DefaultQueryLimit(
+		auth.CheckScopeForRequest("data", http.HandlerFunc(grid.postgRESTHandler)),
```
krilor, line 42: There should be a TODO here to move grid to its own scope.

krilor (dismissed):

PR #958, opened 2026-06-25
feat(ui): make comment visibility clearer and editable (FLEX-1173)

kenneth-strand-statnett:
image
image


kenneth-strand-statnett:
`*_involved_parties` views må utvides til å inkludere flere parties. Hvem skal inkluderes her? 🤔

enzo-crance-statnett (dismissed):

enzo-crance-statnett:
I assume this new visibility has been introduced so that other SOs having to do some prequalifications can read what other SOs have been writing to make the process go faster? So I would expect some of the involved parties views to include something like "all SOs also having a prequalification going on on the same resource" for example.

Needs to be discussed/precised functionally.

local/scripts/templates/comment_rls.j2.sql
```diff
@@ -97,6 +97,32 @@ USING (
     )
 );
 
+DROP POLICY IF EXISTS "{{ data.acronym }}C_SO002_SP002_same_party_type"
+ON {{ resource }}_comment;
+CREATE POLICY "{{ data.acronym }}C_SO002_SP002_same_party_type"
+ON {{ resource }}_comment
+FOR SELECT
+TO flex_system_operator, flex_service_provider
```
krilor: No point in running this policy for SP since it is only really relevant for SO. Removing flex_service_provider saves cycles and improves security IMHO.

local/scripts/templates/comment_rls.j2.sql
```diff
@@ -97,6 +97,32 @@ USING (
...
+        INNER JOIN flex.party AS creator_party
+            ON creator_party.id = comment_creator.party_id
+        INNER JOIN flex.party AS current_party
+            ON current_party.id = (SELECT flex.current_party())
+        WHERE comment_creator.id = {{ resource }}_comment.created_by -- noqa
+            AND creator_party.type = current_party.type
+    )
+    AND EXISTS (
+        SELECT 1
+        FROM flex.{{ resource }}_involved_parties AS {{ lower_acronym }}_ip -- noqa
```
krilor: Using the _involved_parties view doesn't allow for other parties than the involved parties to see comments, I think. I think we need to rethink this policy a bit.

db/flex/service_provider_product_suspension_comment_migrations.sql
```diff
@@ -1,54 +1,27 @@
 --liquibase formatted sql
 -- Manually managed file
 
--- changeset flex:service-provider-product-suspension-comment-visibility-update runOnChange:false endDelimiter:;
+-- changeset flex:service-provider-product-suspension-comment-constraint-typo runOnChange:false endDelimiter:;
```
krilor: 👍

db/flex/service_providing_group_product_application_comment_migrations.sql
```diff
@@ -0,0 +1,18 @@
+--liquibase formatted sql
+-- Manually managed file
```
krilor: I think we should avoid having these ase manually maintained migrations. We should bake this into the code generation in a nicer way.

krilor (dismissed):
Spotted a few things that should be fixed before we can push this further.

krilor:
I have reverted most of changes to visibility originally proposed, but keeping the valuable bits with making visibility levels clearer on the frontend.

Being able to edit the visibility level has also been re-added after it disappeared in a previous refactor. It is important to be able to change the visibility in case of choosing the incorrect one to begin with.

The frontend now looks like this this when editing:

bilde

local/scripts/templates/comment_resource.j2.sql
enzo-crance-statnett, line 1: This migration has no effect since the changes were removed, so it makes noise in the generated files right now, but it gives us the possibility for an easy change across all comment resources later on, so I think it makes sense to keep. 👍🏻

local/scripts/templates/embedding.j2.sql
enzo-crance-statnett, line 1: thx

enzo-crance-statnett (approved):

PR #971, opened 2026-06-30
feat: generated scope enums (FLEX-1175)

enzo-crance-statnett:
This PR makes the list of possible scopes _generated_, both in the DB and in the OpenAPI specification.

Auth scopes and `use` scopes (RPC endpoints) are manual because we don't have an obvious way to read that from a source, and using `resources.yml` is pretty reliable. For the `data` and `grid` module resources, we generate a `read` scope and a `manage` scope based on the possible operations, in such a way that a read-only resource only will have `read`.

This change was the occasion to introduce the `grid` scope asset, and we also removed endpoints related to history tables in the `grid` schema because we don't have history views there so it's only internal and opening the endpoint was a bug.

Regarding migrations, we liberally add `read:grid` to scopes having `read:data` or `manage:data` from before, because the grid resources used to belong to the `data` module API-wise.

We no longer document the whole list of scopes in the Markdown files as it is now way too long. Instead we just give a link to the OpenAPI spec which will contain the exhaustive list.


backend/data/static/openapi.json
```diff
@@ -244,15 +244,106 @@
...
+                    "manage:data:service_providing_group_product_application_comment",
+                    "read:data:service_providing_group_product_application_comment_history",
+                    "read:data:service_providing_group_product_suspension",
+                    "manage:data:service_providing_group_product_suspension",
+                    "read:data:service_providing_group_product_suspension_history",
+                    "read:data:service_providing_group_product_suspension_comment",
+                    "manage:data:service_providing_group_product_suspension_comment",
+                    "read:data:service_providing_group_product_suspension_comment_history",
+                    "read:data:notice",
+                    "use:data:controllable_unit:lookup",
```
krilor, line 353: As per our definition of scopes, the required scope `use:data:controllable_unit:lookup` also results in valid scopes

* `use:data:controllable_unit` <- missing?
* `use:data` <- we have this

db/flex/entity_client_migrations.sql
```diff
@@ -75,3 +37,20 @@ BEFORE INSERT OR UPDATE ON entity_client
 FOR EACH ROW
 WHEN (new.party_id IS NOT null)
 EXECUTE FUNCTION entity_client_check_assumable_party();
+
+-- changeset flex:entity-client-add-grid-scopes runOnChange:false endDelimiter:;
+--preconditions onFail:MARK_RAN
+--precondition-sql-check expectedResult:0 SELECT COUNT(*) FROM flex.entity_client WHERE 'read:grid' = ANY(scopes)
+ALTER TABLE flex.entity_client
```
krilor, line 73: This should ideally be on a session basis. Not really a concern right here since we have slow-moving data and probably no concurrent operations. But doing it this way will disable triggers for all transactions and cause triggers to not run for regular app users.

There are some interesting takes under this SO question:

https://stackoverflow.com/a/72226683

docs/resources/entity_client.md
```diff
@@ -47,18 +47,18 @@ key stored in the entity client.
...
+| <a name="field-party_id" href="#field-party_id">party_id</a>                | Reference to the party this client allows to assume. A null value means the client cannot assume any party.                                                                                                                                                                                                       | bigint                                                                                                                      | [party.id](party.md#field-id)   |
+| <a name="field-scopes" href="#field-scopes">scopes</a>                      | List of scopes granted to the user when it logs in as an entity or when it acts as the party. When assuming a party through party membership, the least privileged set of scopes will be kept.<br/>Scopes are inspired from OAuth 2.0 and allow refinement of access control and privilege delegation mechanisms. | <br/>Required<br/>Array<br/>See [OpenAPI spec](https://elhub.github.io/flex-information-system/api/v1/) for allowed values. |                                 |
```
krilor: Maybe we should consider having a dedicated scopes docs page at some point where we have details about the scopes and put this generated list of scopes.

backend/data/static/openapi.json
```diff
@@ -244,15 +244,106 @@
                 "format": "text",
                 "type": "string",
                 "enum": [
+                    "read:auth",
```
krilor: Havent thought through, but given that we are moving towards multiple services and stuff, would it make sense to enable the use of bare verb scopes `read`, `use`, `manage`?

krilor (commented):
Good stuff.

Though I'm a bit worried about this becoming a place where we (at least technically) frequently introduce breaking changes to the API. In this specific case there is low risk of actually breaking clients - given the assumed interactions with the API. Wish we had a way to say that this is an "open" enum.. Ref https://openapi-code-generator.nahkies.co.nz/guides/concepts/enums

I think the benefit of having enumerated scopes outweighs the downside though. Going to just specifying a pattern makes clients less ergonomic. But up for discussion.

My second worry is what this looks like on the frontend. Do we now have this huge list of scopes as part of party_membership creation?

backend/data/static/openapi.json
enzo-crance-statnett: We now compute all possible prefixes so we have that.

docs/resources/entity_client.md
enzo-crance-statnett: Implemented now ✅

backend/data/static/openapi.json
enzo-crance-statnett: Could be. Not implemented as part of the current change though, as we have the verb/asset semantics assumed a bit everywhere.

enzo-crance-statnett:
Switched to a pattern where scopes are listed in the body of a function, then used as a check constraint on the `flex.scope` domain. This way, we can freely change scopes without having migration issues (only the function changeset will run and all dependent objects will continue to work).

The current state of the PR has been thoroughly tested with respect to migrations, from the last deployed commit in all deployed environments (which are different as we merged the problematic scope changes only in the test environment).

We still lack a frontend change so scopes remain easy to use.

enzo-crance-statnett:
For now the component will remain limited to a practical subset of scopes including the new assets (`grid` and `attachment`). Ideally we should have something better but it will be done as a later task.

Skjermbilde 2026-07-27 102508

kenneth-strand-statnett (changes requested):
Looks like `party_membership_migrations.sql` and `entity_client_migrations.sql` referes to scope `use:attachment` that no longer exists. Replace with `read:attachment`?`

The gin logger seems to have been reintroduced, it should be removed in `flex.go`.

Do we really need to have separate scopes for all of these?
```
        'read:grid',
        'read:grid:line',
        'read:grid:substation',
        'read:grid:substation_cluster',

```
I can't imagine we'll ever need to give access to one and not the other 🤔 


Otherwise it looks good, great work :100:

kenneth-strand-statnett (approved):

PR #974, opened 2026-07-01
chore: split, rewrite and index AP and CU policies for SO to gain speed (FLEX-1326)

krilor:


krilor:
Current suggestion improves response times quite a bit.

Still draft since

* I see a couple failing tests
* Need to evaluate if authorization should be time dependent, i.e. use valid_time_range

krilor:
Spotted the issues and fixed the tests.

Leaving the authorization as non-timedependent for now. That is fairly consistent across all-the-things

enzo-crance-statnett (approved):

PR #984, opened 2026-07-06
feat: attachment resource for SPGPA and basis for automation (FLEX-1323)

enzo-crance-statnett:
This PR shows an example of what an attachment resource can look like in our system, as well as an implementation of the logic for file upload, validation, and storage, with a local S3 mock.

We use Apache PDFBox as a library for validation of PDF files, and the stdlib in Java for images (`ImageIO`). We check structural validity and remove potentially malicious parts, but we do not run a malware detection tool.

The local S3 mock is developed by Adobe so it looked solid enough. But since it's only for local tests the choice of library is less important here. The S3 technology could be abstracted even further if other object storage solutions (Azure etc) work in a similar enough way. As a first version I did not go that far.

The metadata read endpoints are delegated to PostgREST, simply because we can. These just read metadata table rows and not file contents which come with `/download`, so we can use an exposed view like for the other resources. We can also choose to implement our own list and read endpoints in the backend, I just did the simple thing first.


docs-dev/attachments.md
```diff
@@ -0,0 +1,249 @@
...
+authorisation as the metadata read endpoint, as metadata needs to be read before
+the file can be fetched from the storage container.
+
+```http
+GET /res_attachment/19/download
+```
+
+For performance reasons, this endpoint does not directly return the contents of
+the file, but instead uses a redirect mechanism so that the user is sent to an
+URL in the storage container to download the file. Whether the file is opened in
```
krilor: We could be a bit more specific here. E.g. by using the words "bucket" instead of "storage container" and linking to the docs about presigned requests.

We should discuss lifetime (how many seconds the PAR should be valid).

We could also _show_ the responses, instead of just writing about them. To show how we set headers and return codes etc.

docs-dev/attachments.md
```diff
@@ -0,0 +1,249 @@
...
+the browser or downloaded depends on the `Content-Disposition` header that we
+set on redirect. We choose to download by default here.
+
+We do not expose the pointer to the storage container object in the API as it is
+not used in the endpoint paths and can be considered to be an internal
+implementation detail.
+
+### Uploading and deleting files
```
krilor: We should maybe split uploading and deleting into separate sections? And add a section on updates as well (what we don't allow them - just add a new file)?

docs-dev/attachments.md
```diff
@@ -0,0 +1,249 @@
...
+
+The storage container stores data in a key-value format in principle, but we can
+play on the content of these two.
+
+First, the key can be stored in a path-like format so that we do not only have a
+flat logical hierarchy, even though the physical hierarchy will remain flat. We
+can therefore store a key like `/res_attachment/{res_id}/{object_id}` for
+instance.
+
+Second, the value can be stored in a compressed format so network calls go
```
krilor: Didn't we agree that the accepted files (pdf and images) do not need additional compression since they are already heavily optimized?

docs-dev/attachments.md
```diff
@@ -0,0 +1,249 @@
...
+can be trusted to a certain degree.
+
+### Storage container data format
+
+The storage container stores data in a key-value format in principle, but we can
+play on the content of these two.
+
+First, the key can be stored in a path-like format so that we do not only have a
+flat logical hierarchy, even though the physical hierarchy will remain flat. We
+can therefore store a key like `/res_attachment/{res_id}/{object_id}` for
```
krilor: This is where we should include name of the file at the end IMHO. So that browsers being redirected with a PAR will end up at `someoracledomain.com/jibberish/uuid/mysanetizedfilename.png`

enzo-crance-statnett:
List of attachments

Screenshot 2026-07-08 170001

Clicking on delete

Screenshot 2026-07-08 170008

Hovering the file icon shows that the button is clickable (downloads the file)

Screenshot 2026-07-08 170016

enzo-crance-statnett:
- Moved all the attachment backend code to an `attachment` package.
- For more isolation, using a view in the `api` schema instead of raw access to `flex`.
- Kept the API structure, but internally (in `flex`) split the implementation into two tables, one general `attachment` for metadata and one `<res>_attachment` for linking attachments to parent resources.
- Both now have history enabled and we no longer delete files from the bucket.
- Object ID is back in the API because not exposing it makes stuff more complicated when it comes to the view (cannot insert with more fields in the input record than defined on the view).
- The keys in the bucket now have the format `{object_id}/{name}`.
- We assume the multipart form on attachment create comes with parent ID first and then the file contents, so that we can first parse the parent ID and check authorisation _before_ parsing the file contents. Implemented.

We still need to
- add image support
- fix the double traceresponse header bug
- add content disarm and not just parsing
- background worker for deleting dead files (optional for now IMO because files will be dead when they are not mentioned in the history, meaning we have touched it manually -- we can implement the feature when we first decide to prune history)

db/flex/attachment.sql
```diff
@@ -0,0 +1,32 @@
+--liquibase formatted sql
+-- Manually managed file
+
+-- changeset flex:attachment-create runOnChange:true endDelimiter:--
+CREATE TABLE IF NOT EXISTS attachment (
+    id bigint PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+    object_id uuid NOT NULL DEFAULT public.uuid_generate_v4(),
+    name text NOT NULL,
```
krilor: I do think that we need to maintain/store two names.

1. The original filename: Like "my file (énä).pdf" <- returned in content disposition
2. The sanetized filename: Like "my-file-ena.pdf" <- stored in object storage (ref https://docs.oracle.com/en-us/iaas/private-cloud-appliance/cmn/object/managing-storage-objects.htm#managing-storage-objects__object-storage-prefixes-hierarchies - we need to care about both length and characters)

docs-dev/diagrams/attachment_download_sequence.plantuml
```diff
@@ -0,0 +1,51 @@
...
+rnote left of bucket
+""<color #886599>url</color>"", valid for 1 minute
+end note
+deactivate bucket
+
+fis -> user:\
+""301 Moved Permanently""\n\
+Redirect to ""<color #886599>url</color>""
+rnote left of fis
+""Content-Disposition: attachment""
```
krilor: Surprised to see Content-Disposition header on the redirect response. Why do we want that?

docs-dev/attachments.md
```diff
@@ -0,0 +1,290 @@
...
+The PAR request lives for one minute because we estimate this is enough, even
+for a slow client, to receive the URL and start downloading the file.
+
+We do not expose the pointer to the bucket object in the API as it is not used
+in the endpoint paths and can be considered to be an internal implementation
+detail.
+
+### Uploading files
+
+File upload is a non-standard endpoint as it needs to receive the file contents
```
krilor: I think what we want to say is that it does not look like our other endpoints because we want to follow the "general" standard/normal way of handling file uploads?

docs-dev/attachments.md
```diff
@@ -0,0 +1,290 @@
...
+
+### Deleting files
+
+Deleting a file can be done with the delete endpoint with the attachment row ID:
+
+```text
+DELETE /res_attachment/27
+```
+
+Internally, it focuses on deleting the metadata first, then tries to delete from
```
krilor: With history, this changes.

docs-dev/attachments.md
```diff
@@ -0,0 +1,290 @@
...
+the bucket, but this second part is our responsibility and less important for
+the user.
+
+### Updating files
+
+As a first version of the feature, we do not implement file versioning. If a
```
krilor: This points to us wanting to support file update in the future. I think we should not. Maybe we should rephrase to say that updating files is not supported - just upload a new one.

docs-dev/attachments.md
```diff
@@ -0,0 +1,290 @@
...
+
+We keep history of all attachments. There is a functional reason to this: any
+documents that were available to the system operator at the time of the decision
+on an application can lead to a better understanding of the decision. Only very
+old documents should be deleted at some point, but this may just be carried out
+once in a while by administrators if/when the bucket gets too big.
+
+## Storage
+
+In addition to the database and the API endpoints, we need to decide what we
```
krilor: We have made descisions, so should describe them?

docs-dev/attachments.md
```diff
@@ -0,0 +1,290 @@
...
+store in the bucket, in which format exactly, and how we make the link between
+metadata and storage.
+
+### Data quality
+
+When the users upload files, we do not trust the `Content-Type` header and
```
krilor, line 209: I think we should lean on all these three when figuring our what type of file:

* Content-Type
* File extension
* Binary data check

If any of these does not match, we should throw a bad request back to the user.

docs/resources/service_providing_group_product_application_attachment.md
```diff
@@ -0,0 +1,79 @@
...
+
+#### Anonymous
+
+No policies.
+
+#### Common
+
+| Policy key    | Policy                                          | Status |
+|---------------|-------------------------------------------------|--------|
+| SPGPAA-COM001 | Read attachments for SPGPAs that they can read. | DONE   |
```
krilor, line 50: Need to switch this to involved parties only.

enzo-crance-statnett:
Implemented
- image file support
- sanitised filenames
- fixed docs
- fixed `traceresponse` duplicate
- content disarm step for both PDF and images

For content disarm we could have used some libraries, but I thought just looping over the structure and removing code blocks would be enough, and I was not a fan of multiplicating dependencies.

For images it is even simpler as `ImageIO` is a standard Java library in the SDK, and re-encoding apparently strips a lot of potential bad stuff.

For PDF we also had an alternative of taking a screenshot of every page and re-saving as an image. Then the final PDF only has image blocks and nothing else. But this makes stuff heavier I guess, and text can no longer be selected (can be useful for people reading some technical data?).

enzo-crance-statnett:
Skjermbilde 2026-07-16 173423

enzo-crance-statnett:
Tests were checked/corrected and all pass.

Remaining is maybe to move the endpoints to their own DB schema / auth scope (`attachment`)?

kbackend/src/main/kotlin/no/elhub/flex/attachment/AttachmentHandler.kt
```diff
@@ -0,0 +1,275 @@
...
+                .mapLeft { e ->
+                    logger.error { "File validation failed for new attachment: $e" }
+                    when (e) {
+                        is InvalidFileContent ->
+                            ParsingError("Invalid file content: ${e.reason}")
+                    }
+                }
+                .bind()
+
+            // upload to storage
```
kenneth-strand-statnett: I think it would be nice to wrap s3 and db as a transaction to ensure that we don't get stale files in s3.

I believe something like this would work:
```
with(principal) {
    flexTransaction { _ ->
        repo.insert(...).bind()
        attachmentStorageService.upload(...).mapLeft { ... }.bind()
    }
}
```

`repo.insert()` will rollback if s3 fails

kbackend/src/main/kotlin/no/elhub/flex/attachment/AttachmentRepository.kt
```diff
@@ -0,0 +1,209 @@
...
+ * Database implementation of [AttachmentRepository] for a given [baseResource].
+ *
+ * [baseResource] is interpolated directly into SQL to build the view name
+ * (`<baseResource>_attachment`) and its foreign key column (`<baseResource>_id`), since JDBC
+ * can only bind parameter *values*, not identifiers.
+ *
+ * One instance is required per base resource; this class is not a Koin `@Single` because Koin has
+ * no way to supply [baseResource] automatically.
+ */
+class AttachmentRepositoryImpl(private val baseResource: String) : AttachmentRepository {
```
kenneth-strand-statnett, line 89: Maybe add some tests for this repository just to verify that insert, read and delete works as expected on both the attachment table and resource table.

kenneth-strand-statnett (changes requested):
I've added a couple of comments, otherwise it looks good - great work 👏

kbackend/src/main/kotlin/no/elhub/flex/attachment/AttachmentHandler.kt
enzo-crance-statnett: It would be the pattern proposed [there](https://github.com/elhub/flex/pull/1258#discussion_r3517973811).

I was reluctant on the principle of trying anything database-wise before anything is confirmed stored, but as long as we don't commit a DB write before having stored the file it's OK I guess.

We can still have stale files in S3 in any case though. The failure we are talking about would happen while the backend is waiting for the bucket call to succeed. In the existing implementation, it waits without anything in DB. In the proposed change, it waits with a row stored in DB in a non-committed transaction. Committing an open transaction is faster than opening one and storing a row before committing it, so we can possibly avoid some rare cases of backend failures by doing the change. So I think it makes sense even though we are talking about corner cases.

enzo-crance-statnett: Implementing this actually allows fixing a silent bug we have in the transaction handling.

The `flexTransaction` function takes a block returning an `Either`, but if this return value is an error (`Left` case), we do not rollback the transaction. It did not cause issues before because all our transactions probably just do database operations, so if something fails in the repositories it throws and the transaction is rolled back, but here we are exactly in the problematic case where we mix database and non-database operations in a `flexTransaction` block. If the non-database operations fail, the block returns a `Left` value but it is currently not considered a reason for rolling back because no exception was thrown.

kbackend/src/main/kotlin/no/elhub/flex/attachment/AttachmentRepository.kt
enzo-crance-statnett: We don't have much to test here since we already test the behaviour through the API, but you are right it is a bit of a blind spot not to test this trigger.

However, since the repository just makes SQL calls and testing requires reading internal stuff in the DB, I thought it made sense to put the test at the DB level.

kenneth-strand-statnett (approved):

PR #986, opened 2026-07-08
chore: remove null-examples

krilor:
On my fresh dev setup I noticed two things:

1. openapi generation failed on the null example values similar to this: https://github.com/stoplightio/spectral/issues/1981
2. eslint and prettier was in conflict on some formatting stuff in pre-commit

I'm guessing that both of these are due to my setup being with slightly newer versions of the tools. This does not fix that underlying issue, but makes things "work on my machine" ...

### DIFF
--- a/frontend/src/accounting_point/grid_location/AccountingPointGridLocationInput.tsx
+++ b/frontend/src/accounting_point/grid_location/AccountingPointGridLocationInput.tsx
@@ -73,7 +73,7 @@ const GridLocationFormFields = ({
         selectedSubstation.business_id !== currentBusinessId;
       if (isNewSubstation) setValue("nominal_voltage", null);
     }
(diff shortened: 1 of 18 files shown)


frontend/src/accounting_point/grid_location/AccountingPointGridLocationInput.tsx
```diff
@@ -73,7 +73,7 @@ const GridLocationFormFields = ({
         selectedSubstation.business_id !== currentBusinessId;
       if (isNewSubstation) setValue("nominal_voltage", null);
     }
-  }, [selectedSubstation, setValue]);
+  }, [selectedSubstation, currentBusinessId, setValue]);
```
krilor, line 76: eslint was complaining about this missing thing

enzo-crance-statnett (approved):

PR #1017, opened 2026-07-29
refactor(frontend): reusable badge components for SPGPA and SPGPQ

kenneth-strand-statnett:
- Added missing status badge in SPGPA and SPGPQ tables
- Refactored with reusable status badge component


enzo-crance-statnett (approved):
This could maybe be factorised even further by only having one `StatusBadge` parameterised with the resource name. But no need to go that far for now I guess (rule of thumb is that we automate after the third occurrence and there it's only 2 resources needing this).

PR #1019, opened 2026-07-30
feat: set boundary parameter manually in attachment upload multipart form (FLEX-1323)

enzo-crance-statnett:
This PR tries to avoid that our attachment upload requests get wrongly caught by the firewall in the test environment. Indeed, we got errors that did not reach our backend and there is no visible routing error in the deployed cluster, so we assume it comes from the firewall.

After investigating, our hypothesis is that the WAF does not like the `boundary` parameter generated by the browser. It starts with `----` and then random letter/digit characters. Trying with `curl` instead seems to always work, and the generated `boundary` there has way more dashes. We suspect that some mechanism preventing injection attacks does not like the 4 dashes, but with 24 dashes it lets us through.

The proposed fix (even though a better long-term solution would be to actually make the firewall rules more precise and remove those false positives) is to override the way the generated API client writes the attachment upload requests, so that instead of getting a generated `boundary`, we decide what is used there.

### DIFF
--- a/frontend/src/components/attachments/registry.ts
+++ b/frontend/src/components/attachments/registry.ts
@@ -6,6 +6,40 @@ import {
 } from "../../generated-client";
 import { throwOnError } from "../../util";
 
+const enc = new TextEncoder();
+
+// create raw multipart form body
+// used to be able to set the boundary, which is not possible with FormData
+// (the WAF seems not to like the default boundary)
+async function buildMultipartBlob(
+  id: number,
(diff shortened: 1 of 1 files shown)


enzo-crance-statnett:
cc @jonalu14

jonalu14 (dismissed):
Looks ok

jonalu14 (approved):

kenneth-strand-statnett (approved):

PR #1038, opened 2026-08-04
feat(ui): make additional information clearer on SPG create page (FLEX-1200)

enzo-crance-statnett:
Just a bit more space, better text and a placeholder to make it very clear that this is optional.

Skjermbilde 2026-08-04 161509

### DIFF
--- a/frontend/src/service_providing_group/input/ServiceProvidingGroupFields.tsx
+++ b/frontend/src/service_providing_group/input/ServiceProvidingGroupFields.tsx
@@ -1,4 +1,5 @@
 import { useGetIdentity } from "ra-core";
+import { Divider, Heading } from "../../components/ui";
 import { zServiceProvidingGroupCreateRequest } from "../../generated-client/zod.gen";
 import { getFields } from "../../zod";
 import {
@@ -43,10 +44,18 @@ export const ServiceProvidingGroupFields = ({ isEdit }: Props) => {
           description
         />
       )}
+      <div className="pt-6 pb-3">
+        <Divider />
+      </div>
+      <Heading level={4} size="small" className="pb-3">
+        Other
+      </Heading>
       <TextAreaInput
         {...fields.additional_information}
         rows={5}
         description
+        placeholder="This field is optional and can be left empty."
+        descriptionOverride="This field is meant to capture any additional information about the service providing group that might be relevant."
         tooltip={false}
         warning="Please remember not to write any sensitive (power/market/personal) information in this field."
       />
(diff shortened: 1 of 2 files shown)


frontend/src/service_providing_group/input/ServiceProvidingGroupFields.tsx
```diff
@@ -43,10 +44,18 @@ export const ServiceProvidingGroupFields = ({ isEdit }: Props) => {
...
+      </div>
+      <Heading level={4} size="small" className="pb-3">
+        Other
+      </Heading>
       <TextAreaInput
         {...fields.additional_information}
         rows={5}
         description
+        placeholder="This field is optional and can be left empty."
+        descriptionOverride="This field is meant to capture any additional information about the service providing group that might be relevant."
```
simenja: Shouldn't we be supporting multilanguage (en/nn/nb)?

frontend/src/service_providing_group/input/ServiceProvidingGroupFields.tsx
```diff
@@ -43,10 +44,18 @@ export const ServiceProvidingGroupFields = ({ isEdit }: Props) => {
           description
         />
       )}
+      <div className="pt-6 pb-3">
```
simenja, line 48: Are we consistent with our spacings accross the project? Different views on spacing will make different developers make different choices here. 
I am used to have these values as tokens. E.g.
`elhub-spacings-xs`

frontend/src/service_providing_group/input/ServiceProvidingGroupFields.tsx
enzo-crance-statnett: It is not a priority to translate the app at the moment, but we could have an approach similar to the design system migration, where we basically add translations here and there in files we touch, until a good part of the app is done and when we pick up the official internationalisation task, there will be less effort to make.

frontend/src/service_providing_group/input/ServiceProvidingGroupFields.tsx
enzo-crance-statnett: I think we just used numbers everywhere. We actually have the same problem with font size for instance (heading levels).

krilor: Tokens would be nice.

I think the path we have talked about going down is to make better layout/ui components with these things baked in. Establishing patterns through the components. There are a couple of tasks related to this in our backlog that we can talk about prioritising.

frontend/src/service_providing_group/input/ServiceProvidingGroupFields.tsx
krilor: There is multilanguage support that is disabled since we are missing translations for lots of stuff. @kenneth-strand-statnett has been good at adding texts to https://github.com/elhub/flex-information-system/blob/main/frontend/src/intl/text.ts when touching files. I have personally not been good at doing that 🙈

Maybe we can do that here?

krilor (dismissed):

frontend/src/service_providing_group/input/ServiceProvidingGroupFields.tsx
ceccpein: Added translations for the placeholder and descriptionOverride in text.ts

PR #1044, opened 2026-08-06
fix: partydropdown should only list relevant parties by checking party_entity_id and memberships (FLEX-1291)

simenja:


frontend/src/entity/client/EntityClientInput.tsx
```diff
@@ -18,6 +19,25 @@ export const EntityClientInput = () => {
...
+      sort: { field: "id", order: "ASC" },
+      filter: entityId ? { entity_id: entityId } : {},
+    },
+    { enabled: !!entityId },
+  );
+
+  const partiesFromMembershipResource = memberships.map(
+    (m: any) => m.entity_id,
+  );
+  const parties = [...new Set([entityId, ...partiesFromMembershipResource])];
```
enzo-crance-statnett: If I am right you could use an embedding parameter here instead of a second call. Since there is a foreign key from party membership to party, giving `?embed=party` should give you the party record in addition in the response. Check the API specification and maybe search for other places using the `embed` query parameter.

enzo-crance-statnett (dismissed):

frontend/src/entity/client/EntityClientInput.tsx
krilor: I wonder how far we can push the embedding and PostgREST here?
Wonder if we can keep party as the resource, embed membership and filter with or?

If we have a request like this:

`GET ../party?embed=party_membership&or=(entity_id.eq.XXX,party_membership.entity_id.XXXX)`

Shouldn't that get the whole list?

Refs:

* https://elhub.github.io/flex-information-system/api/query-parameters/#embedding-related-resources-embed
* https://docs.postgrest.org/en/v12/references/api/tables_views.html

simenja: @krilor I tried this approach, but keep getting 

`failed to parse logic tree (eq.(entity_id.eq.4,party_membership.entity_id.eq.4))` 

When having this filter
```
    const partyFilter1 = entityId
        ? {
            embed: "party_membership",
            or: `(entity_id.eq.${entityId},party_membership.entity_id.eq.${entityId})`,
        }
        : {}; 
```

krilor (approved):
LGTM

PR #1045, opened 2026-08-06
feat: set trace ID for each synced accounting point (FLEX-1344)

enzo-crance-statnett:
This PR sets a trace ID for each accounting point synced from the adapter. When called from the CU lookup handler, a trace ID already exists and we can use it. When called from the sync job (_i.e._, outside of HTTP handlers), we don't have any trace ID set up and it ends up as `unknown` in our logs which is not very helpful. This is a first step in fixing the sync bug we noticed.

### DIFF
--- a/kbackend/src/main/kotlin/no/elhub/flex/util/TraceIdUtil.kt
+++ b/kbackend/src/main/kotlin/no/elhub/flex/util/TraceIdUtil.kt
@@ -1,5 +1,8 @@
 package no.elhub.flex.util
 
+import kotlinx.coroutines.currentCoroutineContext
+import kotlinx.coroutines.withContext
+import no.elhub.flex.config.TraceContext
 import no.elhub.flex.config.TraceInfo
 import org.slf4j.MDC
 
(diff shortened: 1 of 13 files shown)


kenneth-strand-statnett (changes requested):
Instead of passing TraceInfo through the signatures, we could store it as a CoroutineContext.Element at the call sites and read it via coroutineContext[TraceInfoElement]. That would keep the interfaces clean.
What do you think?

enzo-crance-statnett:
There is something wrong in half-implementing the explicit context-as-argument version and still needing to call `MDC` functions in `traceIdOrUnknown`. Actually we would need a consistent rewrite of the way we store and retrieve trace IDs across the app. My mind was screaming "reader monad" from the first second, but this does not really exist in Kotlin (I keep reading about context receivers being a good enough version of that though). I also don't know exactly what you are referring to. But I can give it more thought and try to find a sweet spot.

kbackend/src/main/kotlin/no/elhub/flex/util/TraceIdUtil.kt
```diff
@@ -1,9 +1,38 @@
...
+ * and with its trace IDs pushed into MDC under `trace_id` and `span_id` for structured logging.
+ *
+ * MDC keys are cleaned up afterwards, even on failure. Intended for non-HTTP contexts
+ * (e.g. scheduled jobs) that have no Ktor call to set up trace context automatically.
+ */
+suspend fun <T> withTrace(traceInfo: TraceInfo, block: suspend () -> T): T {
+    MDC.put("trace_id", traceInfo.traceID)
+    MDC.put("span_id", traceInfo.spanID)
+    try {
+        return withContext(TraceContext(traceInfo)) { block() }
```
kenneth-strand-statnett: I believe we would need to use `kotlinx-coroutines-slf4j` in order to keep correct MDC between suspensions:
`return withContext(TraceContext(traceInfo) + MDCContext()) { block() } `

Ref: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-slf4j/kotlinx.coroutines.slf4j/-m-d-c-context/

kenneth-strand-statnett (changes requested):

kenneth-strand-statnett (approved):

PR #1046, opened 2026-08-07
feat: retry on adapter connection + increased timeout (FLEX-1344)

enzo-crance-statnett:


enzo-crance-statnett:
Closing as we decided to only have increased timeout on FIS for now and rather have retry in the adapter + technology-wise we rely on a standard Ktor plugin only.

PR #1055, opened 2026-08-10
feat(ui): show product type names and MAP (up/down) in SPGPA list (FLEX-1352)

enzo-crance-statnett:
This PR shows the product types as names instead of bare numbers in the SPGPA list and also adds MAP information (up and down directions).

We use the PR as an occasion to replace the `ProductTypeArrayField` component everywhere it is used. I don't know whether we should push it further (_i.e._, migrating to EDS all the concerned pages) but at least the old component is deleted.

Skjermbilde 2026-08-10 095650

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -5400,9 +5400,9 @@
             }
         },
         "node_modules/dompurify": {
-            "version": "3.4.12",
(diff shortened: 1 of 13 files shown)


krilor (approved):
Good stuff. Lets not push it further right now 👍

PR #1067, opened 2026-08-12
fix: filter on active status for spa and spgpa (FLEX-1371)

simenja:


frontend/src/components/ui/formContainer/index.tsx
```diff
@@ -8,7 +8,7 @@ const FormContainer = (props: FormContainerProps) => {
     <Container
       {...props}
       className={cn(
-        "p-3 bg-background flex flex-col gap-5 max-w-7xl mt-4",
+        "p-3 bg-background flex flex-col gap-5 max-w-7xl mt-4 overflow-auto",
```
simenja, line 11: to fix the issue with dropdowns getting out of the form container, couldnt find a better solution atm. Think we need to change the combobox component in EDS

simenja: This issue: 

Image

simenja: the fix:

Image

krilor (approved):

PR #1070, opened 2026-08-13
feat: add view accounting_point_start (FLEX-1351)

krilor:
Purpose is to have a security_definer view that can be used by lookup and sync.

something like `select least(controllable_unit_start_time, controllable_unit_service_provider_valid_time_start) from aps`

### DIFF
--- a/db/flex/accounting_point_start.sql
+++ b/db/flex/accounting_point_start.sql
@@ -1,0 +1,23 @@
+--liquibase formatted sql
+-- Manually managed file
+
+-- changeset flex:accounting-point-start-date-create runOnChange:true endDelimiter:--
+CREATE OR REPLACE VIEW accounting_point_start
+WITH (security_invoker = false) AS (
+    SELECT
+        ap.id AS accounting_point_id,
+        MIN(cu.start_date)::timestamp AT TIME ZONE 'Europe/Oslo'
+            AS controllable_unit_start_time,
+        MIN(LOWER(cusp.valid_time_range))
+            AS controllable_unit_service_provider_valid_time_start
+    FROM flex.accounting_point AS ap
+        LEFT JOIN flex.controllable_unit AS cu
+            ON ap.id = cu.accounting_point_id
+        LEFT JOIN flex.controllable_unit_service_provider AS cusp
+            ON cu.id = cusp.controllable_unit_id
+    GROUP BY ap.id
+);
+
+-- changeset flex:accounting-point-start-date-grants runOnChange:true endDelimiter:;
+GRANT SELECT ON TABLE accounting_point_start
+TO flex_internal_data;
--- a/db/flex/changelog.yml
+++ b/db/flex/changelog.yml
@@ -280,6 +280,9 @@ databaseChangeLog:
       file: ./accounting_point_bidding_zone.sql
       relativeToChangelogFile: true
   - include:
+      file: ./accounting_point_start.sql
+      relativeToChangelogFile: true
+  - include:
       file: ./controllable_unit_bidding_zone.sql
       relativeToChangelogFile: true
   - include:


enzo-crance-statnett:
I am doing the `min` computation in the backend instead of the DB though. IDK which is best, but I thought maybe we might want to log both minimum dates if we get bugs later and we want to investigate.

kenneth-strand-statnett (changes requested):
As discussed, I'd prefer if we do not use database views like this, but instead have the queries directly in the backend. Introducing another layer hides logic, which makes it more difficult to change, test and follow the code. 
I think we should have  a session to decide what to do going forward.

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointRepository.kt
```diff
@@ -159,10 +161,20 @@ interface AccountingPointRepository {
      */
     context(principal: FlexPrincipal)
     suspend fun getByIds(accountingPointIds: List<Long>): Either<RepositoryError, List<AccountingPoint>>
+
```
kenneth-strand-statnett: I don't think the name `getAccountingPointStarts` is very descriptive, compared to the name it was replacing.
Is it really the _starts_ for accounting point or the CUs and service provider "ownership"?

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointRepository.kt
```diff
@@ -487,6 +499,42 @@ class AccountingPointRepositoryImpl : AccountingPointRepository {
...
+    context(principal: FlexPrincipal)
+    override suspend fun getAccountingPointStarts(accountingPointIds: List<Long>): Either<RepositoryError, Map<AccountingPointId, AccountingPointStart>> =
+        flexTransaction { conn ->
+            Either.catch {
+                conn.prepareNamed(
+                    """
+                    SELECT
+                        ap.id AS accounting_point_id,
+                        MIN(cu.start_date)::timestamp AT TIME ZONE 'Europe/Oslo'
```
kenneth-strand-statnett: I think the date to zoned date time conversion should be done in the business logic layer in the kotlin code and not in the query.

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointRepository.kt
```diff
@@ -159,10 +161,20 @@ interface AccountingPointRepository {
      */
     context(principal: FlexPrincipal)
     suspend fun getByIds(accountingPointIds: List<Long>): Either<RepositoryError, List<AccountingPoint>>
+
+    /**
```
kenneth-strand-statnett: Given that we change the name of the function (see other comment), I think this belongs in the repository for Controllable Units. We don't really need to query the accounting_point table at all, do we?

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/AccountingPointService.kt
```diff
@@ -80,8 +81,23 @@ interface AccountingPointService {
...
     context(principal: FlexPrincipal)
     suspend fun getByIds(accountingPointIds: List<Long>): Either<AppError, List<AccountingPoint>>
+
+    // NB: nullable results here because we want to decide at the call site what we do if no data is there
+
+    /**
+     * Gets the earliest date from which we have active data about the given accounting points in the system.
+     */
+    context(principal: FlexPrincipal)
+    suspend fun getAccountingPointStarts(accountingPointIds: List<Long>): Either<AppError, Map<AccountingPointId, Instant?>>
```
kenneth-strand-statnett: See other comment regarding placing this in AP or CU repo.

Also, I don't think the comment is very precise - data about an accounting point does not contain a start date. 
I guess it would be cleaner to have interpretation of data given a context on call site and not in the repositories.

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/AccountingPointService.kt
```diff
@@ -215,6 +231,28 @@ class AccountingPointServiceImpl(
         accountingPointRepository.getByIds(accountingPointIds)
             .mapLeft { InternalServerError(traceIdOrUnknown()) }
 
+    context(principal: FlexPrincipal)
+    override suspend fun getAccountingPointStarts(accountingPointIds: List<Long>): Either<AppError, Map<AccountingPointId, Instant?>> =
+        accountingPointRepository.getAccountingPointStarts(accountingPointIds)
+            .map { apStarts ->
+                apStarts.mapValues {
+                    // earliest of both dates, or null if none exists
+                    if (it.value.controllableUnitStartTime == null && it.value.controllableUnitServiceProviderValidTimeStart == null) {
```
kenneth-strand-statnett: Suggested refactor:

`listOfNotNull(it.value.controllableUnitStartTime, it.value.controllableUnitStartTime).minOrNull()`

kbackend/src/main/kotlin/no/elhub/flex/model/domain/AccountingPointStart.kt
```diff
@@ -0,0 +1,16 @@
...
+import kotlin.time.Instant
+
+/**
+ * Internal domain representation of an accounting point start row from the database.
+ *
+ * @property accountingPointId internal ID of the accounting point
+ * @property controllableUnitStartTime the earliest start date of all CUs behind the accounting point
+ * @property controllableUnitServiceProviderValidTimeStart the earliest start date of all CUSP contracts on CUs behind the accounting point
+ */
+data class AccountingPointStart(
```
kenneth-strand-statnett: See comment regarding naming

kenneth-strand-statnett (changes requested):

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointRepository.kt
enzo-crance-statnett: It is the moment the AP starts to be important in our system, no matter who owns it or what is stored on it. But we could use `getAccountingPointStartDates` or something like that to be a bit clearer.

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/db/AccountingPointRepository.kt
enzo-crance-statnett: How do you know in which order the results come? I wasn't sure the `ANY` filter would preserve order, so getting back something with the input ID was safer.

Note that this is also how other APIs we use tend to work when they take as input a list of IDs and not only a single one (check the end users API for example).

We could maybe select from the given array though, or something like that.

kenneth-strand-statnett: Not sure if I understand your concern, I was thinking something like this:

```
SELECT
    cu.accounting_point_id as accounting_point_id,
    MIN(cu.start_date) as cu_min_start_date,
    MIN(LOWER(cusp.valid_time_range)) as controllable_unit_service_provider_valid_time_start
FROM flex.controllable_unit AS cu
LEFT JOIN flex.controllable_unit_service_provider AS cusp
    ON cu.id = cusp.controllable_unit_id
WHERE cu.accounting_point_id = ANY(:accountingPointIds)
GROUP BY cu.accounting_point_id;
```

kenneth-strand-statnett: Not sure if it would be cleaner to separate the query into two calls/steps, e.g. (dont look at the naming xD)

```
val earliestControllableUnit = repo.getControllableUnitWithEarliestStartDateByAccountingPointId(apId);
val firstServiceProvider = cuspRepo.getEarliest(earliestControllableUnit.id);
val startDate = minOfNotNull(earliestControllableUnit.startDate, firstServiceProvider.valid_time_range.start).minOrNull() ?: LocalDate.now()...
```

krilor:
> As discussed, I'd prefer if we do not use database views like this, but instead have the queries directly in the backend. Introducing another layer hides logic, which makes it more difficult to change, test and follow the code. I think we should have a session to decide what to do going forward.

I agree and disagree. The thing I like about views is that they are nice when looking at data in the DB (which i tend to do). Having premade views into data is a general benefit, tho I do get your point.

kbackend/src/main/kotlin/no/elhub/flex/accountingpoint/AccountingPointService.kt
enzo-crance-statnett: Moved to CU repo because we don't query AP longer.

enzo-crance-statnett:
Leaning towards views in general because by just reloading the DB we know if the thing is correct and we can test it by running manual queries etc. Also agree with putting as much SQL as possible in the DB code instead of scattered around.

But in this specific case we create a view for only one call site, so there is no real factorisation benefit and it's acceptable to have the thing as a simple query. So I won't fight for a view on this even though I think it is generally better.

kenneth-strand-statnett (approved):
I still don't think `getAccountingPointStartDates` is a suitable name for the data it fetches, but otherwise it looks good 💯

PR #1072, opened 2026-08-13
feat: sync MGA no matter their status (FLEX-1381)

enzo-crance-statnett:
This PR makes the MGA sync procedure keep the status of each synchronised MGA. This way, we can store inactive MGAs as well (or not delete MGAs from the system and cause problems when they become inactive).


krilor (dismissed):
Nice

PR #1078, opened 2026-08-14
feat: cu status active when adding to spg (FLEX-1360)

ceccpein:
image


frontend/src/service_providing_group/membership/useAddMembersState.ts
```diff
@@ -107,9 +103,22 @@ export const useAddMembersState = ({ spgId, destination }: Props) => {
     .map((id) => cuById.get(id))
     .filter(Boolean) as CuWithMembership[];
 
+  const isCuSelectable = (cu: CuWithMembership): boolean =>
+    effectiveCheckedIds.has(cu.id) || (cu.status === "active" && cu.isEligible);
+
+  const selectableFilteredCUs = useMemo(
+    () => filteredCUs.filter(isCuSelectable),
+    [filteredCUs, effectiveCheckedIds],
+  );
```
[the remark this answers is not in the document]
ceccpein: I don’t think useMemo is needed for isCuSelectable here. The function is cheap, and the actual expensive work is the filter itself. Memoizing only the filtered result is enough and keeps the code simpler.

backend/data/static/openapi.json
```diff
@@ -2574,7 +2574,7 @@
                 "type": "object",
                 "properties": {
                     "controllable_unit_id": {
-                        "description": "Reference to the controllable unit this relation links to a service providing group. The controllable unit's flexible power must not exceed 100% of the combined maximum active power of all its technical resources (SPGM-VAL003).",
+                        "description": "Reference to the controllable unit this relation links to a service providing group. The controllable unit's flexible power must not exceed 100% of the combined maximum active power of all its technical resources (SPGM-VAL003). The controllable unit must be active before it can be added to a service providing group (SPGM-VAL004).",
```
krilor: I think we should avoid putting all the validation rules in the description. It is documented in the validation rules section so I suggest we remove this. (and the same for VAL003).

The validation rule is not tied to a specific field. That is why it is documented on the resource as a separate thing.

db/flex/service_providing_group_membership.sql
```diff
@@ -172,6 +172,41 @@ BEFORE INSERT OR UPDATE ON service_providing_group_membership
...
+SECURITY INVOKER
+AS $$
+DECLARE
+    lv_status text;
+BEGIN
+    SELECT lower(status::text) INTO lv_status
+    FROM flex.controllable_unit
+    WHERE id = NEW.controllable_unit_id;
+
+    IF lv_status IS NULL THEN
```
krilor: This is enforced somewhere else.

db/flex/service_providing_group_membership.sql
```diff
@@ -172,6 +172,41 @@ BEFORE INSERT OR UPDATE ON service_providing_group_membership
...
+        RAISE 'Controllable unit id=% does not exist', NEW.controllable_unit_id;
+    END IF;
+
+    IF lv_status <> 'active' THEN
```
krilor: We should probably use COALESCE or IS DISTINCT FROM to handle the possible null value

krilor (dismissed):
Good stuff. I've added some non-blocking inline comments

backend/data/static/openapi.json
ceccpein: Good point! Agree, I’ll remove it

db/flex/service_providing_group_membership.sql
```diff
@@ -172,6 +172,37 @@ BEFORE INSERT OR UPDATE ON service_providing_group_membership
...
+    END IF;
+
+    RETURN NEW;
+END;
+$$;
+
+-- changeset flex:service-providing-group-membership-status-check-trigger runOnChange:true endDelimiter:--
+CREATE OR REPLACE TRIGGER
+service_providing_group_membership_status_check
+BEFORE INSERT OR UPDATE ON service_providing_group_membership
```
krilor: This should probably be AFTER INSERT to avoid checking for CUs that don't exist

ceccpein: As discussed, I updated this to BEFORE INSERT and verified that it runs late alphabetically.

